PR #24, opened 2025-03-25
Feat/frontend initial setup and bff

Fardeen802:
Handled the default design for documentation as per the image
Made the dragging transition smooth
Made the documentation looks similar in all places it was used
Added the toggle to minimise and maximise documentation


tarakaswathi-datazip:
@Fardeen802 change the pr name to refactor/documentation_panel or something meaningful

PR #31, opened 2025-04-07
fix: sync StreamPanel checkbox state with selectedStreams

yashmangal112:
- Removed the redundant `checked` local state.
- StreamPanel now directly uses the `isSelected` prop as the source of truth.


tarakaswathi-datazip:
the issue is not resolved yet , when the unselected row is selected it should select the checkbox and selected streams data should be opened on the right 
@yashmangal112 look into this (for more details go through the issue )

PR #78, opened 2025-06-10
fix: dest test connection issue fix

tarakaswathi-datazip:
# Description


server/internal/handlers/destination.go
```diff
@@ -195,6 +195,12 @@ func (c *DestHandler) TestConnection() {
 		return
 	}
 	result, _ := c.tempClient.TestConnection(context.Background(), "destination", "postgres", "latest", req.Config)
+	if result == nil {
+		result = map[string]interface{}{
+			"message": "Connection test failed: Please check your configuration and try again",
```
shubham19may, line 200: what is the reason because of which it failed? needs to mention here right?

PR #114, opened 2025-07-02
feat: run olake cli on already downloaded images

vikash390:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


server/internal/handlers/destination.go
```diff
@@ -184,23 +183,17 @@ func (c *DestHandler) TestConnection() {
...
 	if err != nil {
 		utils.ErrorResponse(&c.Controller, http.StatusInternalServerError, "Failed to encrypt destination config: "+err.Error())
 		return
 	}
-	// TODO: use context provided by request
-	result, err := c.tempClient.TestConnection(context.Background(), "destination", "postgres", "latest", encryptedConfig)
+	driver, version := utils.GetAvailableDriversVersions()
+	if driver == "" || version == "" {
+		driver = "postgres"
+		version = "latest"
```
hash-data, line 194: why updating the driver if version not provided?

server/utils/docker_utils.go
```diff
@@ -33,15 +48,71 @@ func GetDockerHubTags(imageName string) ([]string, error) {
...
+		result[driver] = versions[len(versions)-1]
+	}
+
+	var maxDriver, maxVersion string
+	for driver, version := range result {
+		if version > maxVersion {
+			maxDriver = driver
+			maxVersion = version
+		}
+	}
```
hash-data, line 106: if you just get sorted order then you dont need to do it

server/utils/docker_utils.go
```diff
@@ -33,15 +48,71 @@ func GetDockerHubTags(imageName string) ([]string, error) {
...
 }
+
+// GetAvailableDriversVersions returns the driver with the highest available version locally
+func GetAvailableDriversVersions() (string, string) {
+	drivers := []string{"postgres", "mysql", "oracle", "mongodb"}
+	result := make(map[string]string)
+
+	for _, driver := range drivers {
+		imageName := fmt.Sprintf("olakego/source-%s", driver)
+		versions, err := getLocalImageTags(imageName)
```
hash-data, line 92: can we get local image having prefix olakego/source- , instead of checking just for drivers ?

docker-compose.yml
```diff
@@ -26,6 +26,7 @@ services:
       - "8000:8000" # Expose UI port
       - "8080:8080" # Expose backend API port
     volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
hash-data, line 29: add a comment why it is required?

server/internal/handlers/destination.go
vikash390: it would be fallback in case for first time Olake users, as they won't have any image of any driver, similar to our previous implementation

server/utils/docker_utils.go
vikash390: here i am comparing in between, for each source and that source latest image

server/internal/handlers/destination.go
```diff
@@ -184,23 +183,13 @@ func (c *DestHandler) TestConnection() {
 		utils.ErrorResponse(&c.Controller, http.StatusBadRequest, "Invalid request format")
 		return
 	}
```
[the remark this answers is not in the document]
hash-data: the code got removed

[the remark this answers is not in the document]
hash-data:
```
if req.Type == "" {
		utils.ErrorResponse(&c.Controller, http.StatusBadRequest, "Destination type is required")
		return
	}

	if req.Version == "" {
		utils.ErrorResponse(&c.Controller, http.StatusBadRequest, "Destination version is required")
		return
	}
```

server/internal/handlers/destination.go
hash-data: should not it fail? or ui not handling failiure

server/utils/docker_utils.go
```diff
@@ -33,15 +49,114 @@ func GetDockerHubTags(imageName string) ([]string, error) {
...
+	sort.Slice(tags, func(i, j int) bool {
+		return tags[i] > tags[j] // '>' for descending order
+	})
 
 	return tags, nil
 }
+
+// GetAvailableDriversVersions returns the driver with the highest available version locally
+func GetAvailableDriversVersions(ctx context.Context) (string, string) {
+	images, err := getLocalSourceImages(ctx)
```
hash-data, line 94: should not it be after fallback from docker api?

server/utils/docker_utils.go
```diff
@@ -33,15 +49,114 @@ func GetDockerHubTags(imageName string) ([]string, error) {
...
+		}
+	}
+
+	return maxDriver, maxVersion
+}
+
+// getLocalSourceImages returns all local images with prefix olakego/source-
+func getLocalSourceImages(ctx context.Context) ([]string, error) {
+	return getLocalDockerData(ctx, "olakego/source-")
+}
```
hash-data, line 136: why a new function ?

server/internal/handlers/destination.go
hash-data: let us discuss this some doubts i have

server/internal/telemetry/telemetry.go
```diff
@@ -142,6 +143,9 @@ func getLocationFromIP(ip string) *LocationInfo {
 
 // TrackEvent sends a custom event to Segment
 func TrackEvent(_ context.Context, eventName string, properties map[string]interface{}) error {
```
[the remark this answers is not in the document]
hash-data: Why initialize telemetry when it is disabled? As the track event is called and telemetry gets initialized, the user ID is created, which will be served to the frontend, and the frontend event will get pushed. 
So why create the user file if the user has mentioned to disable telemetry?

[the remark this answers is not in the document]
tarakaswathi-datazip: this disabled check can be added in the inittelemetry itself so as to stop id generation @vikash390  can you make this change , w.r.t to frontend we need to add an seperate env variable using which event tracking can be disabled or that disabled can be sent when user requests for telemetry id , we'll finalize this and will pick this up

PR #119, opened 2025-07-07
fix: send source_type for destination test connection in job flow

tarakaswathi-datazip:
- send source_type in desination test connection in job flow


tarakaswathi-datazip:
this got merged via other branch

PR #129, opened 2025-07-23
refactor: incremental changes 

tarakaswathi-datazip:
- add info tooltips for cursor field and fallback cursor 
- handle fallback removal 
- add a button for fallback cursor on click shows dropdown for fallback selection and added tooltip for that as well
- update default value to fallback cursor

### DIFF
--- a/ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
+++ b/ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
@@ -46,7 +46,7 @@ const StreamConfiguration = ({
 	const [fullLoadFilter, setFullLoadFilter] = useState<boolean>(false)
 	const [partitionRegex, setPartitionRegex] = useState("")
 	const [showFallbackSelector, setShowFallbackSelector] = useState(false)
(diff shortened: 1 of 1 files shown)


tanishaAtDatazip:
update: fallback cursor field should have all available cursor fields

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -623,49 +629,83 @@ const StreamConfiguration = ({
...
+													</label>
+													<Select
+														placeholder="Select default"
+														value={defaultCursorField}
+														onChange={(value: string) => {
+															const newCursorField = value
+																? `${stream.stream.cursor_field}:${value}`
+																: stream.stream.cursor_field
+															stream.stream.cursor_field = newCursorField
+															setDefaultCursorField(value)
```
tanishaAtDatazip, line 666: should we use consistent naming? instead of mixing default and fallback, we should stick to one. fallback would make more sense?

tanishaAtDatazip (approved):

PR #132, opened 2025-07-28
Incremental new requirements 

tarakaswathi-datazip:
- show all operators for all types , remove string based operator filter 
- show available cursor fields in filter dropdown , previously we are showing all fields from schema


tarakaswathi-datazip:
> in case of incorrect cron expression in custom type, job shouldn't be created/edited

handled this by adding a validation before all flows

PR #147, opened 2025-08-13
feat: add automatic ECR login and full image URL support

vikash390:
# Description

Added support for pulling images from both Amazon ECR and Docker Hub, including dynamic default image path construction and fallback to cached images on fetch failure.
Implemented ECR authentication via auth token and integrated it into the image pull workflow.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Tested on an EC2 instance with ECR authentication via auth token integrated into the image pull workflow.

# Screenshots or Recordings


server/internal/docker/runner.go
```diff
@@ -117,8 +117,23 @@ func (r *Runner) ExecuteDockerCommand(ctx context.Context, flag string, command
 }
 
 // buildDockerArgs constructs Docker command arguments
-func (r *Runner) buildDockerArgs(flag string, command Command, sourceType, version, configPath, outputDir string, additionalArgs ...string) []string {
+func (r *Runner) buildDockerArgs(ctx context.Context, flag string, command Command, sourceType, version, configPath, outputDir string, additionalArgs ...string) []string {
 	hostOutputDir := r.getHostOutputDir(outputDir)
+
+	repositoryBase := strings.ToLower(os.Getenv("CONTAINER_REGISTRY_BASE"))
+	imageName := r.GetDockerImageName(sourceType, version)
```
hash-data, line 124: in get docker image name remove empty fallback thing, also remove from all places and add strict check of version.

server/utils/docker_utils.go
```diff
@@ -112,6 +171,16 @@ func GetCachedImages(ctx context.Context) ([]string, error) {
 	return strings.Split(strings.TrimSpace(string(output)), "\n"), nil
 }
 
+// ParseECRDetails extracts account ID, region, and repository name from ECR URI
+func ParseECRDetails(fullImageName string) (accountID, region, repoName string, err error) {
```
hash-data, line 175: add an example how string will look like

server/internal/docker/runner.go
```diff
@@ -117,8 +117,23 @@ func (r *Runner) ExecuteDockerCommand(ctx context.Context, flag string, command
...
+	// If using ECR, ensure login before run
+	if strings.Contains(repositoryBase, "ecr") {
+		imageName = fmt.Sprintf("%s/%s", repositoryBase, imageName)
+		accountID, region, _, err := utils.ParseECRDetails(imageName)
+		if err != nil {
+			logs.Critical("failed to parse ECR details: %s", err)
```
hash-data, line 130: why not returning from here?

server/utils/docker_utils.go
```diff
@@ -120,3 +189,52 @@ func isValidTag(tag string) bool {
...
+		return fmt.Errorf("invalid authorization token format")
+	}
+	username := parts[0]
+	password := parts[1]
+	registryURL := aws.ToString(authData.ProxyEndpoint) // e.g., https://678819669750.dkr.ecr.ap-south-1.amazonaws.com
+
+	// Perform docker login
+	cmd := exec.CommandContext(ctx, "docker", "login", "-u", username, "--password-stdin", registryURL)
+	cmd.Stdin = strings.NewReader(password)
+	output, err := cmd.CombinedOutput()
```
hash-data, line 234: can't you pass password or token directly in command?

server/internal/handlers/destination.go
```diff
@@ -221,8 +222,8 @@ func (c *DestHandler) TestConnection() {
 		}
 		for _, image := range images {
 			if strings.HasSuffix(image, version) {
```
hash-data, line 224: give an example how image will look like and how below method will parse it

server/utils/docker_utils.go
```diff
@@ -23,85 +30,137 @@ type DockerHubTagsResponse struct {
...
+
+	if strings.Contains(repositoryBase, "ecr") {
+		fullImage := fmt.Sprintf("%s/%s", repositoryBase, imageName)
+		tags, err = getECRImageTags(ctx, fullImage)
+	} else {
+		tags, err = getDockerHubImageTags(ctx, imageName)
+	}
+
+	// Fallback to cached if online fetch fails or explicitly requested
+	if err != nil && cachedTags {
```
hash-data, line 47: also log the error in case we have to debug

server/utils/docker_utils.go
```diff
@@ -23,85 +30,137 @@ type DockerHubTagsResponse struct {
...
+func fetchCachedImageTags(ctx context.Context, imageName string) ([]string, error) {
+	repositoryBase := strings.ToLower(os.Getenv("CONTAINER_REGISTRY_BASE"))
 
-		var responseData DockerHubTagsResponse
-		if err := json.NewDecoder(resp.Body).Decode(&responseData); err != nil {
-			return nil, fmt.Errorf("failed to decode Docker Hub response: %s", err)
-		}
+	defaultImage := "olakego/source-"
+	if strings.Contains(repositoryBase, "ecr") {
+		defaultImage = fmt.Sprintf("%s/%s", strings.TrimSuffix(repositoryBase, "/"), defaultImage)
```
hash-data, line 130: can you add comment example how default image looks like after this manipulation

hash-data:
also add env var in makefile

server/internal/docker/runner.go
vikash390: but in case of independent destination creation flow we are not getting source type, if we do not handle fallback , how could we fetch version for that case?

server/utils/docker_utils.go
vikash390: done

vikash390: docker login is persistent, in commend it would fail because daemon has no stored credentials in docker run

hash-data:
```
1. Independent Destination ->
        -> source = "",
        -> ecr repository provided
        result: postgres thrown error, locally availbale none
        -> user reply: i have pushed mysql image then why it is failing, why it is not recommended postgres image also need to be update there
        -> docker, 
        -> same work doing for local need to be done for ecr

2. Destination with job
    a. Create Job 
        i. Use Existing destination 
            -> version, source type
        ii. Create
         -> source , version
    b. Edit Job
        i. Use Existing destination 
        -> version, source type
        ii. Create
        -> source type , version 
```

hash-data:
```
1. Independent Destination ->
        -> source = "",
        -> ecr repository provided
        result: postgres thrown error, locally availbale none
        solution: implementation for local and ecr 

```

server/utils/docker_utils.go
```diff
@@ -23,85 +31,152 @@ type DockerHubTagsResponse struct {
...
 		}
 
-		var tags []string
-		for tag := range tagsMap {
-			tags = append(tags, tag)
+		if err == nil {
+			driverImage = imageName
+			break
 		}
+	}
```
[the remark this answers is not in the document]
hash-data: also how we are making sure that image that is going to be used is of heighest tag as you break in between

server/utils/docker_utils.go
```diff
@@ -112,6 +187,22 @@ func GetCachedImages(ctx context.Context) ([]string, error) {
...
 
+// ParseECRDetails extracts account ID, region, and repository name from ECR URI
+// Example:
+//
+//	Input:  "123456789012.dkr.ecr.us-west-2.amazonaws.com/olakego/source-mysql:latest"
+//	Output: accountID = "123456789012"
+//	        region    = "us-west-2"
+//	        repoName  = "olakego/source-mysql:latest"
+func ParseECRDetails(fullImageName string) (accountID, region, repoName string, err error) {
+	re := regexp.MustCompile(`^(\d+)\.dkr\.ecr\.([a-z0-9-]+)\.amazonaws\.com/(.+)$`)
```
hash-data, line 198: we are sure that pattern will always be same?

server/internal/docker/runner.go
hash-data: it still exist

vikash390: removing

server/utils/docker_utils.go
```diff
@@ -53,16 +55,19 @@ func GetDriverImageTags(ctx context.Context, imageName string, cachedTags bool)
 
 		// Fallback to cached if online fetch fails or explicitly requested
 		if err != nil && cachedTags {
-			logs.Error("failed to fetch image tags online for %s: %v, falling back to cached tags", imageName, err)
-			tags, err = fetchCachedImageTags(ctx, imageName)
+			logs.Error("failed to fetch image tags online for %s: %s, falling back to cached tags", imageName, err)
```
hash-data, line 58: it should be warning

docker-compose.yml
```diff
@@ -15,10 +15,11 @@ x-encryption:
 
 services:
   olake-ui:
-    image: olakego/ui:latest
+    image: ${CONTAINER_REGISTRY_BASE:-registry-1.docker.io}/olakego/ui:latest
```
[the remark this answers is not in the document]
hash-data: i believe we will do it, so add a todo

server/internal/docker/runner.go
```diff
@@ -117,8 +118,29 @@ func (r *Runner) ExecuteDockerCommand(ctx context.Context, flag string, command
...
+			return nil
+		}
+		if err := utils.DockerLoginECR(ctx, region, accountID); err != nil {
+			logs.Critical("failed to login to ECR: %s", err)
+			return nil
+		}
+	}
+
```
hash-data, line 143: remove below latest fallback, as well add a strict check of version in source, destination api

hash-data (approved):

PR #172, opened 2025-09-08
chore: send events to Mixpanel

tarakaswathi-datazip:
- sending events to mixpanel instead of sending it to segment.io
- explicitly send distinct_id and event_original_name


ui/src/api/services/analyticsService.ts
```diff
@@ -11,24 +12,13 @@ const sendAnalyticsEvent = async (
...
-		headers: {
-			"Content-Type": "application/json",
-		},
-		body: JSON.stringify(eventData),
-	})
-
-	if (!response.ok) {
-		throw new Error(`Failed to send analytics event: ${response.statusText}`)
-	}
+	await axios.post(ANALYTICS_ENDPOINT, eventData)
```
vishalm0509, line 15: Not handling error ?

tarakaswathi-datazip: handled error

PR #188, opened 2025-09-16
fix: gap issue spec

tarakaswathi-datazip:


vishalm0509:
There is still no gap

image

tarakaswathi-datazip:
> There is still no gap
> 
> image

Previosuly they were so close , I've added 4px which would be enough if we add more it is applying for all so this is better

PR #189, opened 2025-09-18
fix: telemetry url

ImDoubD-datazip:
# Description
fix related to telemetry url. the TLS is turned true for Insecure Skip Verify.


hash-data:
ci failing

PR #203, opened 2025-09-30
test: :white_check_mark: base setup for automation testing using play…

tarakaswathi-datazip:
- add base setup for automation 
- add flows for testing major cases

### DIFF
--- a/server/tests/test_utils.go
+++ b/server/tests/test_utils.go
@@ -34,6 +34,7 @@ const (
     `
 
 	// Download destination docker-compose
+	// TODO: Either move destination and source config into the same Docker Compose setup or download both from the same location for consistency.
 	downloadDestinationComposeCmd = `
         cd /mnt &&
         curl -fsSL -o docker-compose.destination.yml \
@@ -271,7 +272,7 @@ func ExecCommandWithStreaming(ctx context.Context, t *testing.T, ctr testcontain
 }
 
 // PatchDockerCompose updates olake-ui and temporal-worker to build from local code
-// and prints the patched docker-compose.yml
+// TODO: Remove patch command and find alternative to use local code
 func PatchDockerCompose(ctx context.Context, t *testing.T, ctr testcontainers.Container) error {
 	patchCmd := `
     set -e


ui/playwright.config.ts
```diff
@@ -4,9 +4,9 @@ import { defineConfig, devices } from "@playwright/test"
 
 // Constants
 export const TIMEOUTS = {
-	LONG: 5 * 60 * 1000, // 5 minutes
-	SHORT: 10000, // 10 seconds
-	INDEFINITE: 0, // No timeout (indefinite wait)
+	LONG: 10 * 60 * 1000, // 10 minutes
```
tarakaswathi-datazip: let's have only two long and short with 5 min and 10s

ui/playwright.config.ts
```diff
@@ -12,6 +12,7 @@ export const TIMEOUTS = {
 export default defineConfig({
 	testDir: "./tests",
 	timeout: 10 * 60 * 1000, // 10 minutes global test timeout
+	retries: 2,
```
tarakaswathi-datazip: this will trigger all tests for tests for 2 times when they fails is that fine?

deepanshupal09-datazip: yes.. I added them as a failsafe as sometimes playwright wouldn't click the right thing, I also added some test-ids to prevent that, after adding them that has not happened yet.

ui/tests/pages/CreateSourcePage.ts
```diff
@@ -115,11 +123,20 @@ export class CreateSourcePage extends BasePage {
...
-			.locator("div")
-			.filter({ hasText: new RegExp(`^${connector}$`) })
-			.nth(1)
-			.click()
+
+		await this.page.waitForSelector(".ant-select-dropdown:visible")
+
+		const testId = this.connectorTestIdMap[connector]
+		if (testId) {
+			await this.page.getByTestId(testId).click()
```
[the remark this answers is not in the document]
deepanshupal09-datazip:
```	
private readonly connectorTestIdMap: Record<string, string> = {
		MongoDB: "connector-option-mongodb",
		Postgres: "connector-option-postgres",
		MySQL: "connector-option-mysql",
		Oracle: "connector-option-oracle",
	}
```

this part will be different for `source` and `destination`, so maybe we should keep them here... or we can pass this map in the utils function

wdyt?

ui/playwright.config.ts
deepanshupal09-datazip: maybe we can remove them, but I've still kept them for just in case.

ui/tests/pages/CreateSourcePage.ts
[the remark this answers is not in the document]
tarakaswathi-datazip: remaining part seems to be duplicate right we can move that part

tarakaswathi-datazip:
LGTM!

server/tests/test_utils.go
```diff
@@ -0,0 +1,506 @@
...
+        apt-get install -y docker-compose-plugin &&
+        update-ca-certificates
+    `
+
+	// Download destination docker-compose
+	downloadDestinationComposeCmd = `
+        cd /mnt &&
+        curl -fsSL -o docker-compose.destination.yml \
+            https://raw.githubusercontent.com/datazip-inc/olake/master/destination/iceberg/local-test/docker-compose.yml
+    `
```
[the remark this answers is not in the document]
ImDoubD-datazip: volumes:
      - ./pg_hba.conf:/etc/postgresql/pg_hba.conf
      
This is the problem. we need this wherever we are composing up the postgres docker

PR #232, opened 2025-10-16
feat: append only mode

deepanshupal09-datazip:
# Description

This PR adds append only mode in UI

## Type of change

# How Has This Been Tested?

- [x] Tested append only mode in job edit
- [x] Tested append only mode in job creation
- [x] Tested "All Upsert" and "All Append" in streams

# Screenshots or Recordings

Screenshot 2025-10-16 at 4 06 23 PM

Screenshot 2025-10-16 at 4 06 36 PM

Screenshot 2025-10-16 at 4 06 59 PM

### DIFF
--- a/ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
+++ b/ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
@@ -14,7 +14,6 @@ import {
 import {
 	ColumnsPlusRightIcon,
 	GridFourIcon,
-	InfoIcon,
 	LightningIcon,
 	PlusIcon,
 	SlidersHorizontalIcon,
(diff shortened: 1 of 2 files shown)


deepanshupal09-datazip:
> After saving with `all_append` and reopening `edit_streams`, the tab still shows `all_upsert` even though all streams are already in `append_only`. Just checking is this the intended behavior?
> 
>  Screen.Recording.2025-10-22.at.2.40.38.PM.mov

This switch's default mode is "All Upsert", only when user changes that switch to "All Append", all streams are made to append only mode.

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -729,6 +754,34 @@ const StreamConfiguration = ({
 								</div>
 							)}
 					</div>
+
+					<div
+						className={`mb-4 ${!isSelected ? "font-normal text-gray-500" : "font-medium text-neutral-text"}`}
```
vishalm0509: Can we use `clsx` here for conditional styling. Also check if we can use it anywhere else in this PR

vishalm0509 (approved):
LGTM. Minor comments

shubham19may (approved):

PR #266, opened 2025-11-12
fix: telemetry fixes

deepanshupal09-datazip:
# Description

This PR fixes telemetry issues, changes include:

- [x] add `test_connection_source` and `test_connection_destination` event in UI
- [x] remove the `sources_updated` and `destination_updated` events being sent at wrong place.
- [x] added version in `job_created` event.
- [x] fix the data being sent in `job_created` event

## Type of change

# How Has This Been Tested?

- [x] Tested the new added event for test_connection in mixpanel
- [x] Tested the whole user flow and check if right events are coming in mix panel with all the correct information

# Screenshots or Recordings

### DIFF
--- a/server/internal/handlers/job.go
+++ b/server/internal/handlers/job.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"encoding/json"
 	"fmt"
 	"net/http"
 
@@ -380,14 +379,9 @@ func (h *Handler) GetTaskLogs() {
 
 // @router /internal/worker/callback/sync-telemetry [post]
 func (h *Handler) UpdateSyncTelemetry() {
-	var req struct {
-		JobID      int    `json:"job_id"`
-		WorkflowID string `json:"workflow_id"`
-		Event      string `json:"event"`
-	}
-
-	if err := json.Unmarshal(h.Ctx.Input.RequestBody, &req); err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, "Invalid request format", err)
+	var req dto.UpdateSyncTelemetryRequest
+	if err := UnmarshalAndValidate(h.Ctx.Input.RequestBody, &req); err != nil {
+		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


server/utils/telemetry/job.go
```diff
@@ -27,12 +27,18 @@ func TrackJobCreation(ctx context.Context, job *models.Job) {
...
+			}
 		}
 
 		// Safely add destination properties
 		if job.DestID != nil {
 			properties["destination_type"] = job.DestID.DestType
 			properties["destination_name"] = job.DestID.Name
+			if job.DestID.Version != "" {
+				properties["Destination OLake Version"] = job.DestID.Version
+			}
```
vishalm0509: Here also 👆

ui/src/api/services/destinationService.ts
```diff
@@ -98,17 +100,29 @@ export const destinationService = {
 				//timeout is 0 as test connection takes more time as it needs to connect to the destination
 				{ timeout: 0, disableErrorNotification: true },
 			)
-			return {
+			const result = {
```
vishalm0509: Why new var ? In the catch statement we are not setting variable.

ui/src/api/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const EVENT_TEST_CONNECTION_SOURCE = "test_connection_source"
+export const EVENT_TEST_CONNECTION_DESTINATION = "test_connection_destination"
+export const EVENT_CREATE_JOB_CLICKED = "create_job_clicked"
+export const EVENT_CREATE_SOURCE_CLICKED = "create_source_clicked"
+export const EVENT_CREATE_DESTINATION_CLICKED = "create_destination_clicked"
```
vishalm0509: Can we use enums here or something like this

```ts
export const ANALYTICS_EVENTS = {
  TEST_CONNECTION_SOURCE: "test_connection_source",
  TEST_CONNECTION_DESTINATION: "test_connection_destination",
  CREATE_JOB_CLICKED: "create_job_clicked",
  CREATE_SOURCE_CLICKED: "create_source_clicked",
  CREATE_DESTINATION_CLICKED: "create_destination_clicked",
} as const

export type AnalyticsEvent = typeof ANALYTICS_EVENTS[keyof typeof ANALYTICS_EVENTS]
```

What do you think ? (better or leave it as it is)

vishalm0509 (approved):
LGTM

PR #268, opened 2025-11-18
fix: cursor_field issue and other ui issues

deepanshupal09-datazip:
# Description

This PR fixes several UI issues:

1. default cursor_field not being selected when BFF sends sync mode as "incremental" but has no cursor_field
2. increase timeout for the getJobs, getSources and getDestinations API so that they don't canceled prematurely and show errors
3. fix the issue where sometimes the getClearDestinationStatus API was not being called on Jobs Edit page.
4. fix the issue where cursor still shows as 🚫 for changing Destination DB name in Jobs Edit page.
5. show loading while fetching sources and destinations
6. fix destination table name displayed wrong if user changes the destination db name and switch to some other stream.

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/ui/src/api/services/destinationService.ts
+++ b/ui/src/api/services/destinationService.ts
@@ -6,17 +6,10 @@ import {
 	EntityTestRequest,
 	EntityTestResponse,
 } from "../../types"
-import { getConnectorInLowerCase } from "../../utils/utils"
-
-// TODO: Make it parquet on all places
-const normalizeDestinationType = (type: string): string => {
-	//destination connector typemap
-	const typeMap: Record<string, string> = {
-		"amazon s3": "parquet",
-		"apache iceberg": "iceberg",
-	}
-	return typeMap[type.toLowerCase()] || type.toLowerCase()
-}
+import {
+	getConnectorInLowerCase,
+	normalizeConnectorType,
+} from "../../utils/utils"
 
 export const destinationService = {
 	getDestinations: async () => {
@@ -29,6 +22,7 @@ export const destinationService = {
 				const config = JSON.parse(item.config)
 				return {
 					...item,
+					type: normalizeConnectorType(item.type),
 					config,
 					status: "active",
 				}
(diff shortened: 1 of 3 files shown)


ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -233,16 +235,24 @@ const StreamConfiguration = ({
...
 			}
 		}
 
-		stream.stream.sync_mode = newApiSyncMode
 		onSyncModeChange?.(
 			stream.stream.name,
 			stream.stream.namespace || "",
 			newApiSyncMode,
+			cursorField,
 		)
```
vishalm0509: Do we need to call this in the else block, since in case of `INCREMENTAL` the if block is executed and immediately this block is also executed. Since `SetState` is async, when the second `onSyncModeChange` is called, the `cursorField` can still be undefined sometimes. 


What do you think ?

deepanshupal09-datazip: yes, this should be in separate `else` statement, I will fix that

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -111,22 +114,23 @@ const StreamConfiguration = ({
...
-			const [, defaultField] = stream.stream.cursor_field.split(":")
-			setFallBackCursorField(defaultField)
+		const { fallback: initialFallbackCursor } =
+			getCursorFieldValues(initialCursorField)
+
+		if (initialFallbackCursor) {
+			setFallBackCursorField(initialFallbackCursor)
 			setShowFallbackSelector(true)
 		} else {
 			setFallBackCursorField("")
```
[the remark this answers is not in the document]
vishalm0509: So do we need to set both the times or just once in the outer block ?
```ts
setFallBackCursorField(initialFallbackCursor)
setFallBackCursorField("")
```

vishalm0509 (approved):
LGTM

PR #273, opened 2025-11-21
chore: guard clear-destination by source version and enforce unique names for jobs/sources/destinations

vishalm0509:
# Description

### **guard clear-destination by source version and enforce unique names for jobs/sources/destinations**

- add a reusable version-compatibility check and use it inside the clear-destination and stream-difference flows so older connectors cannot perform clear-destination or edit-streams.
- expose a single `/project/:projectId/check-unique` API that validates job/source/destination names.
- wire the new endpoint into the UI so source/destination names are checked before submission
- document the new contract, remove the legacy handler/route, and clean up unused DTO/service code

Fixes (None)

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [x] Scenario A
exercised job creation, source creation (standalone + job flow), destination creation, stream difference modal
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/ui/src/modules/jobs/pages/JobCreation.tsx
+++ b/ui/src/modules/jobs/pages/JobCreation.tsx
@@ -322,7 +322,7 @@ const JobCreation: React.FC = () => {
 					ENTITY_TYPES.JOB,
 				)
 				if (!isUnique) return
-				
+
 				//TODO : Job name is disabled once filled and moved to next step , need to be handled later
 				setJobNameFilled(true)
 				setCurrentStep(JOB_CREATION_STEPS.SOURCE)


server/internal/database/job.go
```diff
@@ -173,14 +174,34 @@ func (db *Database) DeleteJob(id int) error {
...
+
+// IsSourceNameUniqueInProject checks if a source name is unique within a project.
+func (db *Database) IsSourceNameUniqueInProject(ctx context.Context, projectID, name string) (bool, error) {
+	return db.IsNameUniqueInProject(ctx, projectID, name, constants.SourceTable)
+}
+
+// IsDestinationNameUniqueInProject checks if a destination name is unique within a project.
+func (db *Database) IsDestinationNameUniqueInProject(ctx context.Context, projectID, name string) (bool, error) {
+	return db.IsNameUniqueInProject(ctx, projectID, name, constants.DestinationTable)
+}
```
vikaxsh: move these to respective files

server/internal/services/etl/utils.go
```diff
@@ -190,3 +192,12 @@ func waitForSyncToStop(ctx context.Context, tempClient *temporal.Temporal, proje
 
 	return nil
 }
+
+// checks the version compatibitlity for clear-destination and stream difference operation
```
vikaxsh:
```suggestion
// checks the version compatibility for clear-destination and stream difference operation
```

ui/src/api/services/destinationService.ts
```diff
@@ -151,4 +151,19 @@ export const destinationService = {
...
 		return response.data
 	},
+
+	checkDestinationNameUnique: async (
+		destinationName: string,
+	): Promise<{ unique: boolean }> => {
+		try {
+			const response = await api.post<{ unique: boolean }>(
+				`${API_CONFIG.ENDPOINTS.PROJECT(API_CONFIG.PROJECT_ID)}/check-unique`,
+				{ name: destinationName, entity_type: "destination" },
```
deepanshupal09-datazip:
```suggestion
				{ name: destinationName, entity_type: ENTITY_TYPES.DESTINATION },
```

Similar to what you did in source and destination service.

ui/src/modules/jobs/pages/JobCreation.tsx
deepanshupal09-datazip: > this is not needed as we can directly show the message returned by backend, just set the `showNotification` flag to be true in the service layer while calling the API like
> 
> ```
> 			const response = await api.post<{ unique: boolean }>(
> 				`${API_CONFIG.ENDPOINTS.PROJECT(API_CONFIG.PROJECT_ID)}/check-unique`,
> 				{ name: destinationName, entity_type: ENTITY_TYPES.DESTINATION },
> 				{ showNotification: true },
> 			)
> ```
> 
> please change this wherever necessary.



please make this change here too.

ui/src/api/services/destinationService.ts
vishalm0509: Done

ui/src/modules/destinations/pages/CreateDestination.tsx
```diff
@@ -378,6 +380,21 @@ const CreateDestination = forwardRef<
...
+			if (isUnique === null) {
+				message.error("Failed to verify destination name uniqueness")
+				return
+			}
+			if (!isUnique) {
+				message.error(
+					"Destination name already exists. Please choose a different name.",
+				)
+				return
+			}
```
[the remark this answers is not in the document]
vishalm0509: In this case it will show success toast also right. We need to show only error toast messages for this API.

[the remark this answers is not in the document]
deepanshupal09-datazip: We should also display the success message returned by the backend, as we do for other API calls. This will keep the behavior consistent across the app.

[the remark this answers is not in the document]
vishalm0509: I mean unique check is internal, we should not show success message for it right. We can only show if it fails. We are not showing for job also.

vikaxsh (approved):

PR #294, opened 2025-12-27
fix: pass null state in case of manual clear destination

vishalm0509:
# Description

- Updated `etl.ClearDestination` to determine if it's manual `(syncWaitTime == 0)` or edit-triggered
- For manual clear-destination, save `job.State = "{}"` to database before triggering workflow
- Worker continues to read state from database (no changes required)

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/server/internal/services/etl/job.go
+++ b/server/internal/services/etl/job.go
@@ -265,7 +265,7 @@ func (s *ETLService) ActivateJob(ctx context.Context, jobID int, req dto.JobStat
 	return nil
 }
 
-func (s *ETLService) ClearDestination(ctx context.Context, projectID string, jobID int, streamsConfig string, syncWaitTime time.Duration, fullClear bool) error {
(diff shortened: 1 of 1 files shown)


server/internal/services/etl/job.go
```diff
@@ -294,6 +294,14 @@ func (s *ETLService) ClearDestination(ctx context.Context, projectID string, job
 		return fmt.Errorf("failed to wait for sync to stop: %s", err)
 	}
 
+	// for manual clear-destination, update the state file to empty object
+	if fullClear {
```
hash-data:
```suggestion
	if emptyState {
```

hash-data (approved):

server/internal/services/etl/job.go
vishalm0509: Changed to `resetState`

PR #304, opened 2026-01-21
fix: prevent auto-selecting latest version on source edit page

deepanshupal09-datazip:
# Description

This PR fixes a bug where navigating to the Source "Edit" page would incorrectly display the latest available version instead of the actual configured version.

The issue was caused by a race condition in the useEffect hook responsible for fetching available versions. During client-side navigation, the effect would sometimes run before the source data was fully loaded (stale state), causing the "auto-select version" logic to trigger wrongly and overwrite the correct version.

Since the connector type is locked in Edit mode, the auto-select logic was unnecessary. This fix simply removes that logic from `SourceEdit.tsx`and `DestinationEdit.tsx`, ensuring the saved version is always respected.

## Type of change

# How Has This Been Tested?

- [x] Scenario A: Verified that navigating from the Sources/Destinations list to the Edit page correctly preserves and displays the saved version (e.g., v0.3.13) instead of defaulting to the latest.
- [x] Scenario B: Verified that refreshing the Edit page continues to display the correct version.
- [x] Scnario C: Verified that in Job Edit page versions of Source/Destinations are correctly displayed 

# Screenshots or Recordings

### DIFF
--- a/ui/src/modules/destinations/pages/DestinationEdit.tsx
+++ b/ui/src/modules/destinations/pages/DestinationEdit.tsx
@@ -217,14 +217,6 @@ const DestinationEdit: React.FC<DestinationEditProps> = ({
 
 				if (response?.version) {
 					setVersions(response.version)
-
-					// If no version is selected, set the first one as default
-					if (!selectedVersion && response.version.length > 0) {
-						setSelectedVersion(response.version[0])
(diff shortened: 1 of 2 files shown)


vishalm0509 (approved):
LGTM

PR #316, opened 2026-02-10
fix: select default cursor field on init

deepanshupal09-datazip:
# Description

This PR fixes issue when initializing StreamConfiguration, if the default sync mode is 'incremental' and no cursor field is selected, automatically default to the first available cursor field.  

## Type of change

# How Has This Been Tested?

- [x] Tested when default sync mode is "incremental" a default cursor field is selected.

# Screenshots or Recordings


vishalm0509:
Shouldn't the default one be a key ? Right now it's random. Also the test still fails

image

PR #321, opened 2026-02-13
feat: support custom driver version for local testing of olake changes

vishalm0509:
# Description

The goal of this PR is to bridge the gap between development on the olake core (CLI) and the olake-ui. It introduces a `CUSTOM_DRIVER_VERSION` configuration that allows developers to build local driver images and use them immediately in the UI without semantic versioning restrictions.

- **Local Image Injection**: The value of `CUSTOM_DRIVER_VERSION` is automatically appended to the versions list for all connectors, making it selectable from the UI.
- **Semver Bypass**: Compatibility gates are bypassed when a custom version is active.
- **Workflow Refactoring**: Centralized the check logic to ensure developers don't have to manually update version across the codebase when testing new CLI changes or spec/ui-schema changes.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Related PR's (If Any):
https://github.com/datazip-inc/olake-docs/pull/351


server/utils/docker_utils.go
```diff
@@ -301,3 +306,21 @@ func DockerLoginECR(ctx context.Context, region, registryID string) error {
 
 	return nil
 }
+
+// --- developer utils ---
+
+// CustomDriverVersion returns the custom driver version used to test olake with olake-ui.
+// Note: This is only for development/testing purposes.
+// When a custom version is set, semver-based compatibility checks will bypassed.
+func CustomDriverVersion() string {
```
deepanshupal09-datazip:
```suggestion
func GetCustomDriverVersion() string {
```

deepanshupal09-datazip (commented):
I was thinking about how we can simplify the configuration a bit more. Since we already use `runmode = localdev` for testing local UI changes, what if we use that existing runmode check instead of introducing `APP_ENV`?

We could also just specify `CUSTOM_DRIVER_VERSION` directly in `app.conf`. This way, developers wouldn't have to manage two separate environment modes, and all the local configuration stays in one central place.

What do you think about this approach?

server/utils/docker_utils.go
```diff
@@ -301,3 +307,21 @@ func DockerLoginECR(ctx context.Context, region, registryID string) error {
...
 	return nil
 }
+
+// --- developer utils ---
+
+// CustomDriverVersion returns the custom driver version used to test olake with olake-ui.
+// Note: This is only for development/testing purposes.
+// When a custom version is set, semver-based compatibility checks will bypassed.
+func GetCustomDriverVersion() string {
+	if strings.EqualFold(GetAppEnv(), constants.AppEnvDevelopment) {
```
deepanshupal09-datazip: we are already normalizing the string, why do we need `strings.EqualFold` then?

vishalm0509: It won't break anything, but fine. Made changes

deepanshupal09-datazip (approved):
LGTM

PR #327, opened 2026-02-25
feat: add filter_config in selected_streams

deepanshupal09-datazip:
# Description

Refactored the filter payload strategy to support a new structured `filter_config` JSON field for jobs on source versions >= v0.6.0, while fully preserving the existing legacy filter string for older jobs.

- Uses `filter_config` only when source version >= v0.6.0 and no selected stream already carries a non-empty legacy `filter` string
- Existing jobs with a legacy `filter` string stay on the legacy path unconditionally, even if the source version is >= v0.6.0
- Add `filter_config` conditon value validation and type conversions according to column types
- show namespace and stream name too in error.

## Type of change

# How Has This Been Tested?

- [x] Tested for old jobs, `filter_config` is not present
- [x] Tested for old job with source version upgraded, `filter_config` is not present. 
- [x] Tested for new jobs created with source version >= 0.6.0, `filter_config` is present, and `filter` is not being sent.
- [x] Tested for new jobs in edit flow, `filter_config` is being correctly passed in the payload 

# Screenshots or Recordings


ui/src/types/streamTypes.ts
```diff
@@ -12,9 +12,15 @@ export type FilterOperator = "=" | "!=" | ">" | "<" | ">=" | "<="
...
 
 export type FilterCondition = {
-	columnName: string
+	column: string
 	operator: FilterOperator
-	value: string
+	value: string | null
+}
+
+export interface FilterConfigCondition {
```
vishalm0509: Why `FilterCondition` and `FilterConfigCondition`. Can we think of a way to use the same ?

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
 	}
 }
 
-// Returns a copy of the selected streams map with all disabled streams removed
+// Casts a single filter condition's value to the appropriate native type
+// (number, boolean, object, array) based on the column's type schema.
+export const castFilterConditionValue = (
+	cond: FilterConfigCondition,
+	columnSchema?: { type: string | string[] },
```
vishalm0509: When can the type be a `string`. Even for one type it's an array right.

Example:
```json
    "FirstName": {
                            "type": [
                                "string"
                            ],
                            "destination_column_name": "firstname"
                        }
```

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
...
+		case "integer_small":
+		case "integer":
+			castValue = castValue === "" ? null : parseInt(castValue, 10)
+			break
+		case "number_small":
+		case "number":
+			castValue = castValue === "" ? null : parseFloat(castValue)
+			break
+		case "boolean":
+			castValue = castValue === "true"
```
vishalm0509: What if the user enters `TRUE`, `True`

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
...
+	Object.keys(selectedStreams).forEach(key => {
+		result[key] = selectedStreams[key].filter(stream => !stream.disabled)
+	})
+
+	return result
+}
+
+// Applies type casting to filter_config values for their correct native types
+export const formatSelectedStreamsPayload = (
+	selectedStreams: { [key: string]: SelectedStream[] },
```
vishalm0509: We are using this type in multiple places right, can we create a type for it ?

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
...
+): { [key: string]: SelectedStream[] } => {
+	const filteredStreams = getSelectedStreams(selectedStreams)
+
+	const typeSchemaByName = new Map(
+		streams?.map(s => [s.stream.name, s.stream.type_schema?.properties]) ?? [],
+	)
+
 	return Object.fromEntries(
-		Object.entries(selectedStreams).map(([key, streams]) => [
+		Object.entries(filteredStreams).map(([key, nsStreams]) => [
```
vishalm0509: what is `nsStream`, can we name it more explicit ?

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
...
+
+	return result
+}
+
+// Applies type casting to filter_config values for their correct native types
+export const formatSelectedStreamsPayload = (
+	selectedStreams: { [key: string]: SelectedStream[] },
+	streams?: StreamData[],
+): { [key: string]: SelectedStream[] } => {
+	const filteredStreams = getSelectedStreams(selectedStreams)
```
vishalm0509: Why is this `filteredStreams` we aren't filtering anything right

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
...
+
+	const typeSchemaByName = new Map(
+		streams?.map(s => [s.stream.name, s.stream.type_schema?.properties]) ?? [],
+	)
```
vishalm0509: There can be same streamName across different namespaces. So can we format key as `namespace.streamName` this is how we use ID (key) in OLake CLI as well.

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+): boolean => {
+	const typeArray = Array.isArray(type) ? type : [type]
+	if (typeArray.length === 0) return false
+
+	const v = value === null ? "" : String(value)
+
+	return typeArray.some(t => {
+		switch (t) {
+			case "null":
+				return value === null || v === "<null>"
```
vishalm0509: Please check when can `type` be just `string`.

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+
+			case "array": {
+				try {
+					return Array.isArray(JSON.parse(v))
+				} catch {
+					return false
+				}
+			}
+
+			case "object": {
```
vishalm0509: Do we need to handle `TRUE/FALSE/true/false`, did you discuss with @vikaxsh

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+// Validates a structured filter_config object.
+// Returns null if valid, or a descriptive error string.
+export const validateFilterConfig = (
+	filterConfig: FilterConfig,
+	streamName: string,
+	namespace: string,
+	typeSchemaProperties?: Record<string, { type: string | string[] }>,
+): string | null => {
+	const streamPrefix = `[${namespace}.${streamName}]`
+
```
vishalm0509: though this case won't happen, we need to fail here right

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+	return typeArray.some(t => {
+		switch (t) {
+			case "null":
+				return value === null || v === "<null>"
+
+			case "integer_small":
+			case "integer":
+				return /^-?\d+$/.test(v) && Number.isInteger(Number(v))
+
```
vishalm0509:
```suggestion
	const raw = value === null ? "" : String(value).trim()
```

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+	for (const cond of filterConfig.conditions) {
+		if (typeof cond.column !== "string" || cond.column.trim() === "") {
+			return `${streamPrefix} Filter condition is missing a column`
+		}
+		// Values can be null if the schema allows it, but they cannot be missing/undefined entirely
+		if (cond.value === undefined) {
+			return `${streamPrefix} Filter condition for "${cond.column}" is missing a value`
+		}
+
+		// Type-aware validation when schema is available
```
vishalm0509: Instead of this can we return Error if there's an error ?

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+export const validateStreams = (
+	selections: { [key: string]: SelectedStream[] },
+	streams?: StreamData[],
+): string | null => {
+	// Map typeSchemaProperties by stream name for quick lookup
+	const typeSchemaByName = new Map(
+		streams?.map(s => [s.stream.name, s.stream.type_schema?.properties]) ?? [],
 	)
+
+	for (const [namespace, nsStreams] of Object.entries(selections)) {
```
vishalm0509: this part has high potential of refactor. 

```ts
export const validateFilterConfig = (
	filterConfig: FilterConfig,
	streamName: string,
	namespace: string,
	typeSchemaProperties?: Record<string, { type: string | string[] }>,
): string | null => {
	const streamPrefix = `[${namespace ? `${namespace}.` : ""}${streamName}]`

	if (!filterConfig.conditions || filterConfig.conditions.length === 0) {
		return `${streamPrefix} Filter conditions cannot be empty`
	}

	for (const cond of filterConfig.conditions) {
		if (typeof cond.column !== "string" || cond.column.trim() === "") {
			return `${streamPrefix} Filter condition is missing a column`
		}

		// Value can be null, but must not be undefined.
		if (cond.value === undefined) {
			return `${streamPrefix} Filter condition for "${cond.column}" is missing a value`
		}

		if (!typeSchemaProperties) continue

		const columnSchema = typeSchemaProperties[cond.column]
		if (!columnSchema) continue

		const typeArray = Array.isArray(columnSchema.type)
			? columnSchema.type
			: [columnSchema.type]

		if (isValueValidForTypes(cond.value, typeArray)) continue

		const nonNullTypes = typeArray.filter(t => t !== "null")
		const expectedTypes = (nonNullTypes.length ? nonNullTypes : ["null"]).join(" | ")

		return `${streamPrefix} Invalid value "${cond.value}" for column "${cond.column}" — expected type: ${expectedTypes}`
	}

	return null
}
```

--- 

Here is an example for reference

ui/src/utils/utils.ts
```diff
@@ -649,14 +652,88 @@ export const handleSpecResponse = (
 	}
 }
 
-// Returns a copy of the selected streams map with all disabled streams removed
+// Casts a single filter condition's value to the appropriate native type
+// (number, boolean, object, array) based on the column's type schema.
+export const castFilterConditionValue = (
```
vishalm0509: There's a lot of util for filter now. I am thinking of creating a `utils/filter.go`. What do you say ?

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+
+// Returns null if all stream filters are valid, or a descriptive error string.
+export const validateStreams = (
+	selections: { [key: string]: SelectedStream[] },
+	streams?: StreamData[],
+): string | null => {
+	// Map typeSchemaProperties by stream name for quick lookup
+	const typeSchemaByName = new Map(
+		streams?.map(s => [s.stream.name, s.stream.type_schema?.properties]) ?? [],
 	)
```
vishalm0509: same here. `namespace.streamName`

ui/src/utils/utils.ts
```diff
@@ -667,12 +744,144 @@ export const validateFilter = (filter: string): boolean => {
...
+				).join(" | ")
+				return `${streamPrefix} Invalid value "${cond.value}" for column "${cond.column}" — expected type: ${expectedTypes}`
+			}
+		}
+	}
+
+	return null
+}
+
+// Returns null if all stream filters are valid, or a descriptive error string.
```
vishalm0509: This function is a general validator right, not just for filter. So we need to update the comment accordingly. Also there's the same kinda comment in another function please check

ui/src/modules/jobs/pages/JobEdit.tsx
```diff
@@ -361,10 +365,12 @@ const JobEdit: React.FC = () => {
 			return
 		}
 
-		if (
-			!validateStreams(getSelectedStreams(selectedStreams.selected_streams))
-		) {
-			message.error("Filter Value cannot be empty")
+		const filterError = validateStreams(
```
vishalm0509: validateStreams is general right, so having a var name as  `filterError` is misleading

ui/src/modules/jobs/pages/JobEdit.tsx
```diff
@@ -373,8 +379,9 @@ const JobEdit: React.FC = () => {
 				jobId,
 				JSON.stringify({
 					...selectedStreams,
-					selected_streams: getSelectedStreams(
-						selectedStreams.selected_streams,
+					selected_streams: formatSelectedStreamsPayload(
+						getSelectedStreams(selectedStreams.selected_streams),
+						selectedStreams.streams,
 					),
```
vishalm0509: We are using `formatSelectStreamsPayload` and `getSelectedStreams` which both returns the same and used in multiple places. Can we use once instead where we format also. Now there's a confusion as to when to use what. If there's one, it will be simple. Think about it

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -61,14 +62,17 @@ const StreamConfiguration = ({
...
 	onFullLoadFilterChange,
+	onFilterConfigChange,
 	fromJobEditFlow = false,
 	initialSelectedStreams,
 	destinationType = DESTINATION_INTERNAL_TYPES.S3,
 	onIngestionModeChange,
 	onSelectedColumnChange,
 	sourceType,
 }: ExtendedStreamConfigurationProps) => {
+	const useFilterConfig = typeof onFilterConfigChange === "function"
```
vishalm0509: can we check if not `undefined` here ?

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -194,7 +198,26 @@ const StreamConfiguration = ({
...
 		}
-		// Parse initial filter if exists
+
+		if (useFilterConfig) {
+			if (initialFilterConfig && initialFilterConfig.conditions.length > 0) {
+				setMultiFilterCondition({
+					conditions: initialFilterConfig.conditions,
+					logicalOperator: initialFilterConfig.logical_operator,
+				})
+				setFullLoadFilter(true)
```
vishalm0509: Why is it named `fullLoadFilter` ? Now filter is for All modes

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
```diff
@@ -365,20 +388,28 @@ const StreamConfiguration = ({
 		setMultiFilterCondition({
 			conditions: [
 				{
-					columnName: "",
+					column: "",
 					operator: "=",
-					value: "",
+					value: null,
```
vishalm0509: in legacy we didn't allow `null` (staging). Now we are allowing, have you checked it doesn't break anything and backward compatibility is maintained ?

vishalm0509: - legacy (filter string): use ""
- new (filter_config): allow null
And ensure legacy string building never runs with null.

ui/src/modules/jobs/utils/streams.ts
```diff
@@ -114,3 +115,18 @@ export function isColumnEnabled(
...
+export function shouldUseFilterConfig(
+	selectedStreams: SelectedStreamsByNamespace,
+	sourceVersion: string,
+): boolean {
+	if (!sourceVersion || !semver.valid(sourceVersion)) return false
+	if (!semver.gte(sourceVersion, MIN_JSON_FILTER_VERSION)) return false
+
+	// If ANY stream already carries a legacy filter string, keep legacy path.
+	return !Object.values(selectedStreams).some(streams =>
+		streams.some(s => typeof s.filter === "string" && s.filter !== ""),
```
vishalm0509:
```suggestion
		streams.some(s => typeof s.filter === "string" && s.filter.trim() !== ""),
```

ui/src/utils/utils.ts
deepanshupal09-datazip: we are filtering out disabled streams, so that we dont try to typecast filter config value of disabled streams.

ui/src/utils/utils.ts
deepanshupal09-datazip: it is handled by `default` case, because anything entered in a textbox is `string`. if you want I can add `case: "string"` for better readibility.

ui/src/utils/utils.ts
deepanshupal09-datazip: which case?

ui/src/utils/utils.ts
deepanshupal09-datazip: you mean like `new Error()`?
we already return error message if error and `null` if no error

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
deepanshupal09-datazip: yes, the filter_regex makes sure that empty value is not passed

ui/src/types/streamTypes.ts
deepanshupal09-datazip: done, removed `FilterCondition` now.

ui/src/utils/utils.ts
deepanshupal09-datazip: using `toLowerCase()` now.

ui/src/utils/utils.ts
deepanshupal09-datazip: done, using existing `SelectedStreamsByNamespace`.

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: done, using `namespace.steamName` format now

ui/src/utils/utils.ts
deepanshupal09-datazip: done, it can be case insensitive.

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: done, moved to `utils/filterUtils.ts`

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/modules/jobs/pages/JobEdit.tsx
deepanshupal09-datazip: done

ui/src/modules/jobs/pages/JobEdit.tsx
deepanshupal09-datazip: now using only `formatSelectStreamsPayload` in components, `getSelectedStreams` is now being used internally in the utils function itself.

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
deepanshupal09-datazip: done

ui/src/modules/jobs/pages/streams/StreamConfiguration.tsx
deepanshupal09-datazip: changed to `filter`

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/modules/jobs/utils/streams.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
```diff
@@ -649,30 +654,91 @@ export const handleSpecResponse = (
...
+	const result: SelectedStreamsByNamespace = {}
+
+	Object.keys(selectedStreams).forEach(key => {
+		result[key] = selectedStreams[key].filter(stream => !stream.disabled)
+	})
+
+	return result
+}
+
+// Applies type casting to filter_config values for their correct native types
```
vishalm0509: This is a generic utils and not just for filter, can we generalize the comment

ui/src/utils/utils.ts
```diff
@@ -649,30 +654,91 @@ export const handleSpecResponse = (
 	}
 }
 
-// Returns a copy of the selected streams map with all disabled streams removed
-export const getSelectedStreams = (selectedStreams: {
-	[key: string]: SelectedStream[]
-}): { [key: string]: SelectedStream[] } => {
+// Filters out disabled streams
+export const getSelectedStreams = (
```
vishalm0509: Don't export it if it's meant to be internal to utils.

ui/src/utils/filterUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+	if (!filter.trim()) return false
+	return FILTER_REGEX.test(filter.trim())
+}
+
+// Validates if value is compatible with any given DataType.
+// Explicitly handles "null" values by checking if the column schema allows it.
+// Converts other values to string internally to handle native types safely.
+export const isValueValidForTypes = (
+	value: any,
+	type: string | string[],
```
vishalm0509: type will always be a `[]string` not just string.

ui/src/utils/utils.ts
vishalm0509: If error, why do we return string, shouldn't we return `Error`

ui/src/utils/utils.ts
vishalm0509: It will always be a `[]string`

ui/src/utils/utils.ts
deepanshupal09-datazip: we can, but then we will have to add `try catch` blocks wherever we use it, this approach looks more cleaner to me and for a validation check, I think its fine.

ui/src/utils/filterUtils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: done

ui/src/utils/utils.ts
deepanshupal09-datazip: removed `string`

vishalm0509 (dismissed):

ui/src/modules/ingestion/features/jobs/utils/streams.ts
```diff
@@ -221,3 +289,18 @@ export const getCursorFieldValues = (
 		fallback: fallback || "",
 	}
 }
+
+// Returns true if filter_config (JSON) should be used instead of the legacy filter string.
+// Requires source >= v0.4.1 AND no selected stream already carries a non-empty legacy filter.
```
vishalm0509: We need to update the comment to the correct source

deepanshupal09-datazip: done

vishalm0509 (approved):

PR #333, opened 2026-03-02
feat: fusion

badalprasadsingh:
# Description

Adding our new product of compaction on OLake UI. Basically, all its API integrations.

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B


server/conf/app.conf
```diff
@@ -16,3 +16,7 @@ logsdir = ${LOGS_DIR||./logger/logs}
 sessionon = ${SESSION_ON||true}
 TEMPORAL_ADDRESS = ${TEMPORAL_ADDRESS||temporal:7233}
 CONTAINER_REGISTRY_BASE = ${CONTAINER_REGISTRY_BASE||registry-1.docker.io}
+ENABLE_COMPACTION = ${ENABLE_COMPACTION||false}
```
hash-data:
```suggestion
ENABLE_FUSION = ${ENABLE_FUSION||false}
```

server/conf/app.conf
```diff
@@ -16,3 +16,7 @@ logsdir = ${LOGS_DIR||./logger/logs}
...
+COMPACTION_BASE_URL = ${COMPACTION_BASE_URL||http://127.0.0.1:1630}
```
hash-data: please use fusion instead of compaction

server/internal/constants/constants.go
```diff
@@ -50,6 +50,11 @@ var (
 	ConfDeploymentMode        = "DEPLOYMENT_MODE"
 	ConfRunMode               = "runmode"
 	ConfContainerRegistryBase = "CONTAINER_REGISTRY_BASE"
+	ConfEnableCompaction      = "ENABLE_COMPACTION"
+	ConfCompactionBaseURL     = "COMPACTION_BASE_URL"
+	ConfCompactionUsername    = "COMPACTION_USERNAME"
+	ConfCompactionPassword    = "COMPACTION_PASSWORD"
```
hash-data: we can remove username and password, this will be same as olake-ui, it will help us in rbac

server/internal/services/temporal/client.go
```diff
@@ -40,7 +40,7 @@ func NewClient() (*Temporal, error) {
 			taskQueue: constants.TemporalTaskQueue,
 		}
 		return nil
-	}, 3, time.Second)
+	}, 5, 2*time.Second)
```
hash-data: any reason ?

server/go.mod
```diff
@@ -21,6 +22,7 @@ require (
 	go.temporal.io/sdk v1.39.0
 	golang.org/x/crypto v0.47.0
 	golang.org/x/mod v0.32.0
+	google.golang.org/api v0.265.0
```
hash-data, line 25: any reason of moving these?

server/internal/services/etl/services.go
```diff
@@ -2,25 +2,28 @@ package etl
...
 	client, err := temporal.NewClient()
 	if err != nil {
 		return nil, err
 	}
 
 	return &Service{
-		db:       db,
-		temporal: client,
+		db:         db,
+		temporal:   client,
```
hash-data: why this should get to know about compaction client?

hash-data: services are always encapsulated

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+	Catalog      *catalog.Service
+	Table        *table.Service
+	Optimization *optimization.Service
+	Aggregator   *aggregator.Service
+}
+
+func InitService(db *database.Database, client *client.Compaction) (*Service, error) {
+	catalogSvc := catalog.NewService(client)
+	tableSvc := table.NewService(client)
+
```
hash-data: what is diff in both ?

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+		db:           db,
+		client:       client,
+		Catalog:      catalogSvc,
+		Table:        tableSvc,
+		Optimization: optimization.NewService(client, tableSvc),
+		Aggregator:   aggregator.NewService(client, catalogSvc, tableSvc),
+	}
+
+	// Sync all existing destinations at startup
+	logger.Info("Syncing all destinations with compaction service...")
```
hash-data: let us move it inside the function

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *Service) SyncAllDestinations(ctx context.Context) {
+	destinations, err := s.db.ListDestinations()
+	if err != nil {
+		logger.Errorf("Failed to list destinations for compaction sync: %v", err)
+		return
+	}
+
+	catalogSvc := catalog.NewService(s.client)
+	for _, dest := range destinations {
+		if !strings.EqualFold(dest.DestType, "iceberg") {
```
hash-data: let us add a debug log on skip of catalog

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+			logger.Errorf("Failed to sync catalog %s for destination %s: %v", catalogName, dest.Name, err)
+		} else {
+			logger.Infof("Synced catalog %s for destination %s", catalogName, dest.Name)
+		}
+	}
+}
+
+// extractCatalogNameFromConfig extracts the catalog_name field from the config JSON
+func extractCatalogNameFromConfig(configJSON string) (string, error) {
+	var config models.Config
```
hash-data: let us not build any model for it

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+
+	if config.CatalogName == "" {
+		return nil, fmt.Errorf("catalog_name is required in config")
+	}
+
+	catalogType := normalizeCatalogType(string(config.CatalogType))
+	compactionReq := &models.CatalogRequest{
+		Name:            config.CatalogName,
+		Type:            catalogType,
+		OptimizerGroup:  models.OptimizerGroup,
```
hash-data: it will be env variable

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+		TableFormatList: []string{"ICEBERG"},
+		StorageConfig:   make(map[string]string),
+		AuthConfig:      make(map[string]string),
+		Properties:      make(map[string]string),
+		TableProperties: make(map[string]string),
+	}
+
+	compactionReq.StorageConfig["storage.type"] = models.DefaultStroageType
```
hash-data: does it going to work for all ? i read the comment to make sure we have checked it properly

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// TODO: test for each and every catalog and map the keys to respsective values
+func mapCatalogProperties(olakeConfig *models.Config, properties map[string]string, olakeCatalogType string) {
+	warehouse := olakeConfig.IcebergS3Path
+
+	switch strings.ToLower(olakeCatalogType) {
+	case "glue":
+		properties["warehouse"] = warehouse
+
```
hash-data: let us give a thought how we can avoid this handling, replacing keys or making sure olake also accept the same?

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+			logger.Errorf("Failed to extract catalog_name from destination %s config: %v", dest.Name, err)
+			continue
+		}
+
+		exists, err := catalogSvc.CheckCatalogExists(ctx, catalogName)
+		if err != nil {
+			logger.Warnf("Failed to check catalog existence for %s: %v", catalogName, err)
+		}
+
+		if err := catalogSvc.SyncCatalogToFusion(ctx, dest.Config, exists); err != nil {
```
hash-data: it will always update with latest destination, or should we take destination in descending order of created at ?

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+			logger.Infof("Synced catalog %s for destination %s", catalogName, dest.Name)
+		}
+	}
+}
+
+// extractCatalogNameFromConfig extracts the catalog_name field from the config JSON
+func extractCatalogNameFromConfig(configJSON string) (string, error) {
+	var config models.Config
+	if err := json.Unmarshal([]byte(configJSON), &config); err != nil {
+		return "", fmt.Errorf("failed to parse config: %w", err)
```
hash-data: use same error formatting everywhere

server/internal/constants/constants.go
hash-data: I have sent env variable names please use those

server/internal/services/compaction/resources/catalog/sync.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/datazip-inc/olake-ui/server/internal/services/compaction/models"
+	"github.com/datazip-inc/olake-ui/server/utils/logger"
+)
+
+func (s *Service) SyncCatalogToFusion(ctx context.Context, configJSON string, isUpdate bool) error {
+	CompactionCatalogReq, err := MapETLConfigToCompactionCatalog(configJSON)
```
hash-data: Use proper function names  like MapDestToCatalog

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/datazip-inc/olake-ui/server/utils"
+)
+
+func MapETLConfigToCompactionCatalog(configJSON string) (*models.CatalogRequest, error) {
+	var config models.Config
+	if err := json.Unmarshal([]byte(configJSON), &config); err != nil {
+		return nil, fmt.Errorf("failed to parse OLake config: %w", err)
+	}
+
+	if config.CatalogName == "" {
```
hash-data: unmarshalling second time

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+		return "custom"
+	}
+}
+
+func mapAuthConfig(olakeConfig *models.Config, authConfig, cmpStorageConfig map[string]string) {
+	utils.SetIfNotEmpty(cmpStorageConfig, "storage.s3.region", olakeConfig.Region)
+	utils.SetIfNotEmpty(cmpStorageConfig, "storage.s3.endpoint", olakeConfig.S3Endpoint)
+
+	if olakeConfig.AccessKey != "" && olakeConfig.SecretKey != "" {
+		authConfig["auth.type"] = "AK/SK"
```
hash-data: make these constants

server/internal/services/optimization/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+	if _, exists := req.Properties["table.self-optimizing.enabled"]; !exists {
+		req.Properties["table.self-optimizing.enabled"] = "false"
+	}
+	if _, exists := req.Properties["table.self-optimizing.quota"]; !exists {
+		req.Properties["table.self-optimizing.quota"] = "0.1"
+	}
+	if _, exists := req.Properties["cache-enabled"]; !exists {
+		req.Properties["cache-enabled"] = "false"
+	}
+	if _, exists := req.Properties["created-at"]; !exists {
```
hash-data: everytime on update as well it will be added ?

server/internal/services/compaction/resources/catalog/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+	if config.CatalogName == "" {
+		return nil, fmt.Errorf("catalog_name is required in config")
+	}
+
+	catalogType := normalizeCatalogType(string(config.CatalogType))
+	compactionReq := &models.CatalogRequest{
+		Name:            config.CatalogName,
+		Type:            catalogType,
+		OptimizerGroup:  models.OptimizerGroup,
+		TableFormatList: []string{"ICEBERG"},
```
hash-data: use constants

server/internal/services/optimisation/mapper.go
```diff
@@ -0,0 +1,227 @@
...
+		return models.CatalogType(compactionType)
+	}
+}
+
+// setDefaultCatalogProperties sets required default properties for all catalogs
+func setDefaultCatalogProperties(req *models.CatalogRequest) {
+	if req.Properties == nil {
+		req.Properties = make(map[string]string)
+	}
+
```
hash-data: first time map initialization

server/internal/services/compaction/resources/catalog/sync.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to map OLake config to Compaction catalog: %w", err)
+	}
+
+	// Mark this catalog as created from ETL
+	if CompactionCatalogReq.Properties == nil {
+		CompactionCatalogReq.Properties = make(map[string]string)
+	}
```
hash-data: second time initialization

server/internal/services/compaction/resources/catalog/sync.go
```diff
@@ -0,0 +1,101 @@
...
+	CompactionCatalogReq.Properties["olake_created"] = "true"
```
hash-data: CompactionCatalogReq -> newOptimizeCatalogConfig

hash-data: we can use optimize keyword only everywhere instead of fusion and compaction

hash-data: also this property for?

server/internal/services/compaction/resources/catalog/sync.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		logger.Warnf("Failed to check if catalog exists in Compaction: %v", err)
+	}
+
+	if !catalogExists {
+		return nil
+	}
+
+	resp, err := s.UpdateCatalog(ctx, req.Name, req)
+	if err != nil {
```
hash-data: just hit the request

server/internal/services/compaction/resources/catalog/sync.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+
+	if !resp.Success {
```
hash-data: status code check?

server/internal/services/compaction/services.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/datazip-inc/olake-ui/server/internal/services/compaction/resources/catalog"
+	"github.com/datazip-inc/olake-ui/server/internal/services/compaction/resources/optimization"
+	"github.com/datazip-inc/olake-ui/server/internal/services/compaction/resources/table"
+	aggregator "github.com/datazip-inc/olake-ui/server/internal/services/compaction/views"
+	"github.com/datazip-inc/olake-ui/server/utils/logger"
+)
+
+// Service is a unified service for compaction operations
+type Service struct {
+	db           *database.Database
```
hash-data: let us make sure things which are not required not available in this service

server/routes/router.go
```diff
@@ -114,4 +114,33 @@ func Init(h *handlers.Handler) {
...
+		// cron
+		web.Router("/api/v1/fusion/tables/:catalog/:database/:table/cron", compactionHandler, "put:SetCompactionCronConfig")
+		web.Router("/api/v1/fusion/tables/:catalog/:database/:table/cron", compactionHandler, "get:GetCompactionCronConfig")
+
+		// runs
+		web.Router("/api/v1/fusion/tables/:catalog/:database/:table/runs", compactionHandler, "get:GetCompactionRuns")
+		web.Router("/api/v1/fusion/tables/:catalog/:database/:table/cancel", compactionHandler, "post:CancelCompactionProcess")
+
+		// Compaction table properties routes
+		web.Router("/api/v1/fusion/tables/:catalog/:database/:table/enable-optimizing", compactionHandler, "post:EnableSelfOptimizing")
```
hash-data: why we have these much api entered here?

hash-data (commented):
Handler not reviewed yet, as api removal required

server/internal/constants/constants.go
badalprasadsingh: We cannot use OLake's "admin", "password" here. Let's discuss it.

As discussed, keeping it "OPTIMISATION" not "FUSION", as Fusion is not a generic name.

server/internal/constants/constants.go
```diff
@@ -50,6 +50,12 @@ var (
 	ConfDeploymentMode        = "DEPLOYMENT_MODE"
 	ConfRunMode               = "runmode"
 	ConfContainerRegistryBase = "CONTAINER_REGISTRY_BASE"
+	ConfEnableOptimisation    = "ENABLE_OPTIMISATION"
+	ConfOptimisationBaseURL   = "OPTIMISATION_BASE_URL"
+	ConfOptimisationUsername  = "OPTIMISATION_USERNAME"
+	ConfOptimisationPassword  = "OPTIMISATION_PASSWORD"
```
hash-data: we still have these?

server/internal/services/optimisation/models/models.go
```diff
@@ -0,0 +1,52 @@
+package models
```
hash-data: models have a folder currently why we are creating two same package in one project?

server/internal/handlers/optimisation/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+package constants
```
hash-data: why separate constants

server/routes/router.go
```diff
@@ -109,9 +109,28 @@ func Init(h *handlers.Handler) {
...
+	if h.Optimisation != nil {
+		optHandler := h.Optimisation
+
+		// catalogs: crud, (delete in piggy-backing)
+		web.Router("/api/ams/v1/catalog", optHandler, "post:CreateCatalog")
+		web.Router("/api/ams/v1/catalog/:catalog", optHandler, "get:GetCatalog")
+		web.Router("/api/ams/v1/catalog/:catalog", optHandler, "put:UpdateCatalog")
+
+		// terminal: cron, enable/disable optimisation
+		web.Router("/api/ams/v1/:catalog/:database/:table/set", optHandler, "put:SetProperties")
```
hash-data:
```suggestion
		web.Router("/api/ams/v1/:catalog/:database/:table/config", optHandler, "put:SetProperties")
```

server/internal/services/services.go
```diff
@@ -17,12 +25,52 @@ func InitAppService(db *database.Database) (*AppService, error) {
...
 		etl: etlSvc,
-	}, nil
+		opt: optSvc,
+	}
+
+	if enableOptimisation {
+		appSvc.CreateAllDestAsCatalogs(context.Background())
+	}
+
+	return appSvc, nil
```
hash-data:
```suggestion
	appSvc := &AppService{
		db:  db,
		etl: etlSvc,
		opt: nil,
	}

    enableOptimisation := web.AppConfig.DefaultBool(constants.ConfEnableOptimisation, false)

    if enableOptimisation {
		optSvc, err = optimisation.InitService()
		if err != nil {
			return nil, err
		}
        appSvc.opt = optSvc

		appSvc.CreateAllDestAsCatalogs(context.Background())
	}

	return appSvc, nil
```

server/internal/services/services.go
```diff
@@ -17,12 +25,52 @@ func InitAppService(db *database.Database) (*AppService, error) {
...
 }
+
+func (s *AppService) Optimisation() *optimisation.Service {
+	return s.opt
+}
+
+func (s *AppService) CreateAllDestAsCatalogs(ctx context.Context) {
+	destinations, err := s.db.ListDestinations()
+	if err != nil {
+		logger.Errorf("Failed to list destinations for optimisation sync: %v", err)
```
hash-data: sometime using %s and sometime %v

server/internal/services/services.go
```diff
@@ -17,12 +25,52 @@ func InitAppService(db *database.Database) (*AppService, error) {
...
+
+	appSvc := &AppService{
 		db:  db,
 		etl: etlSvc,
-	}, nil
+		opt: optSvc,
+	}
+
+	if enableOptimisation {
+		appSvc.CreateAllDestAsCatalogs(context.Background())
```
hash-data:
```suggestion
		appSvc.SyncCatalog(context.Background())
```

hash-data: Also it is going to re create destination to catalog even we have created them via api

server/internal/handlers/handler.go
```diff
@@ -1,18 +1,120 @@
...
+	Optimisation *optimisation.Handler
 }
 
 func NewHandler(appSvc *services.AppService) *Handler {
-	return &Handler{
+	h := &Handler{
 		ETL: etl.NewHandler(appSvc.ETL()),
 	}
+	if appSvc.Optimisation() != nil {
+		fmt.Println("optimisation is on")
```
hash-data: remove fmt prints

server/internal/services/services.go
```diff
@@ -17,12 +25,52 @@ func InitAppService(db *database.Database) (*AppService, error) {
...
+
+	for _, dest := range destinations {
+		if !strings.EqualFold(dest.DestType, "iceberg") {
+			continue
+		}
+
+		if _, err := s.opt.CreateCatalogFromOLakeConfig(ctx, dest.Config); err != nil {
+			logger.Errorf("Failed to create catalog: %s", err)
+		} else {
+			logger.Infof("Catalog created successfully")
```
hash-data: add log about catalog created from which dest

hash-data: also which catalog created

server/internal/services/optimisation/catalog.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	if config.CatalogName == "" {
+		return "", fmt.Errorf("catalog_name not found in config")
+	}
+
+	return config.CatalogName, nil
+}
+
+func MapETLConfigTooptimisationCatalog(configJSON string) (*cmpModels.CatalogRequest, error) {
```
hash-data:
```suggestion
func CreateOptimizationConfig(configJSON string) (*cmpModels.CatalogRequest, error) {
```

server/internal/services/optimization/table.go
```diff
@@ -0,0 +1,223 @@
...
+
+		healthScore, ok := tableSummary["healthScore"].(float64)
+		if !ok {
+			return nil, fmt.Errorf("missing or invalid healthScore in tableSummary for table %s.%s.%s", catalog, databaseName, tableName)
+		}
+
+		tableInfo.HealthScore = int(healthScore)
+
+		// fetch latest optimizing processes for each type
+		res, err := s.fetchLatestProcessInfo(ctx, catalog, databaseName, tableName, "MINOR")
```
hash-data, line 100: why not calling it with all ?

server/internal/handlers/handler.go
```diff
@@ -1,18 +1,120 @@
...
 	"github.com/datazip-inc/olake-ui/server/internal/handlers/etl"
+	"github.com/datazip-inc/olake-ui/server/internal/handlers/optimisation"
+	"github.com/datazip-inc/olake-ui/server/internal/models/dto"
 	"github.com/datazip-inc/olake-ui/server/internal/services"
+	optService "github.com/datazip-inc/olake-ui/server/internal/services/optimisation"
+	"github.com/datazip-inc/olake-ui/server/utils"
+	"github.com/datazip-inc/olake-ui/server/utils/logger"
 )
 
 type Handler struct {
```
hash-data, line 12: why we are creating in this way let us discuss it

server/internal/services/optimization/table.go
badalprasadsingh: cause anyways we need to extract only the latest compaction process for each type. If we do it will all, we need to iterate to get the latest for each type.

server/internal/services/services.go
badalprasadsingh: This was discussed right, if we create a destination in etl -> create the subsequent catalog in fusion

server/internal/services/optimisation/models/models.go
badalprasadsingh: because these models were only specific to "optimisation" anyways, I have moved them to global models.

server/internal/constants/constants.go
badalprasadsingh: discussed

server/internal/services/compaction/resources/catalog/mapper.go
badalprasadsingh: discussed

server/internal/services/compaction/services.go
badalprasadsingh: done

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,196 @@
...
+	if catalogName == "" {
+		return nil, fmt.Errorf("catalog name is required")
+	}
+
+	path := fmt.Sprintf("%scatalogs/%s", constants.OptimizationAPIBase, catalogName)
+	var result dto.CatalogRequest
+	if err := s.DoInto(ctx, http.MethodGet, path, url.Values{}, nil, &result); err != nil {
+		return nil, fmt.Errorf("failed to get catalog %s: %s", catalogName, err)
+	}
+
```
hash-data, line 35: we must send same status code received from fusion

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,126 @@
...
+		return
+	}
+
+	utils.SuccessResponse(&h.Controller, result.Message, nil)
+}
+
+func (h *Handler) requiredCatalog() (string, bool) {
+	catalog := h.Ctx.Input.Param(":catalog")
+	if catalog == "" {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "catalog name is required", nil)
```
hash-data:
```suggestion
		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "catalog name not present in query params", nil)
```

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,126 @@
...
+
+const badRequestStatusCode = http.StatusBadRequest
+
+func (h *Handler) GetCatalog() {
+	catalogName, ok := h.requiredCatalog()
+	if !ok {
+		return
+	}
+
+	logger.Debugf("Get catalog details initiated catalog[%s]", catalogName)
```
hash-data: let us not add any log here

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,196 @@
...
+	return s.UpdateCatalog(ctx, catalogReq.Name, catalogReq)
+}
+
+func (s *Service) GetCatalogAsOLakeConfig(ctx context.Context, catalogName string) (*models.Config, error) {
+	catalog, err := s.GetCatalog(ctx, catalogName)
+	if err != nil {
+		return nil, err
+	}
+
+	return MapoptimizationCatalogToOLakeConfig(catalog)
```
hash-data:
```suggestion
	return MapCatalogToDest(catalog)
```

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,126 @@
...
+	if !h.bindJSON(&req) {
+		return
+	}
+
+	if req == nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "catalog config is required", nil)
+		return
+	}
+
+	logger.Debugf("Create catalog initiated")
```
hash-data: remove logs from each handler

server/utils/utils.go
```diff
@@ -715,6 +715,20 @@ func GetLogArchiveFilename(jobID int, filePath string) (string, error) {
 	return filename, nil
 }
 
+func ToJSON(v interface{}) (string, error) {
```
hash-data:
```suggestion
func MarshalToString(v interface{}) (string, error) {
```
or can remove the function

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Convert config to JSON string
+	configJSON, err := utils.ToJSON(req)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "invalid config format", err)
+		return
+	}
+
+	result, err := h.opt.CreateCatalogFromOLakeConfig(h.Ctx.Request.Context(), configJSON, false)
```
hash-data:
```suggestion
	result, err := h.opt.CreateCatalog(h.Ctx.Request.Context(), configJSON, false)
```

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,126 @@
...
+	logger.Debugf("Update catalog initiated catalog[%s]", catalogName)
+
+	// Convert config to JSON string
+	configJSON, err := utils.ToJSON(req)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "invalid config format", err)
+		return
+	}
+
+	result, err := h.opt.UpdateCatalogFromOLakeConfig(h.Ctx.Request.Context(), configJSON)
```
hash-data:
```suggestion
	result, err := h.opt.UpdateCatalog(h.Ctx.Request.Context(), configJSON)
```

server/internal/services/optimization/client.go
```diff
@@ -0,0 +1,252 @@
...
+	// If no valid parameters, return current date in yyyyMMdd format
+	if len(parts) == 0 {
+		return time.Now().UTC().Format("20060102")
+	}
+
+	return strings.Join(parts, "")
+}
+
+func (c *Service) sendRequest(ctx context.Context, method, path string, queryParams url.Values, bodyBytes []byte) ([]byte, int, http.Header, error) {
+	signature := c.calculateSignature(queryParams)
```
hash-data: do we need to calculate it everytime?

server/internal/handlers/optimization/piggyBacking.go
```diff
@@ -0,0 +1,75 @@
...
+	utils.SuccessResponse(&h.Controller, "request forwarded successfully", upstreamResponse)
+}
+
+func transformOptPathToAMS(path string) string {
+	const optPrefix = "/api/opt/v1/"
+	const amsPrefix = "/api/ams/v1/"
+
+	if len(path) >= len(optPrefix) && path[:len(optPrefix)] == optPrefix {
+		return amsPrefix + path[len(optPrefix):]
+	}
```
hash-data: why not direct replace logic ?

server/internal/handlers/optimization/piggyBacking.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		body = json.RawMessage(raw)
+	}
+
+	transformedPath := transformOptPathToAMS(req.URL.Path)
+
+	data, headers, err := h.opt.ProxyWithHeaders(req.Context(), req.Method, transformedPath, req.URL.Query(), body)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, http.StatusBadGateway, "upstream request failed", err)
```
hash-data: pass exact same status code back

server/internal/services/optimization/table.go
```diff
@@ -0,0 +1,225 @@
...
+	"net/url"
+	"time"
+
+	"github.com/datazip-inc/olake-ui/server/internal/constants"
+	"github.com/datazip-inc/olake-ui/server/internal/models/dto"
+)
+
+// GetTablesWithDetails fetches all tables with full details for a specific catalog and database
+func (s *Service) GetTablesWithDetails(ctx context.Context, catalog, databaseName string) (*dto.TablesResponse, error) {
+	response := &dto.TablesResponse{
```
hash-data, line 19: just to mention we usually not create responses in services, services are for main logic

server/internal/services/optimization/table.go
```diff
@@ -0,0 +1,225 @@
...
+
+		response.Tables = append(response.Tables, tableInfo)
+	}
+
+	return response, nil
+}
+
+// GetTables returns the list of tables for a given catalog and database
+func (s *Service) GetTables(ctx context.Context, catalog, database, keywords string) (interface{}, error) {
+	path := fmt.Sprintf("%scatalogs/%s/databases/%s/tables", constants.OptimizationAPIBase, catalog, database)
```
hash-data: make these request endpoint constant

server/internal/services/optimization/table.go
```diff
@@ -0,0 +1,225 @@
...
+			if hours == 1 {
+				lastRun = "1 hour ago"
+			} else {
+				lastRun = fmt.Sprintf("%d hours ago", hours)
+			}
+		default:
+			days := int(duration.Hours() / 24)
+			if days == 1 {
+				lastRun = "1 day ago"
+			} else {
```
hash-data: why we are doing this we just need to pass the time ui itself will handle this

server/internal/services/optimization/terminal.go
```diff
@@ -0,0 +1,141 @@
...
+		properties["self-optimizing.minor.trigger.interval"] = config.MinorTriggerInterval
+	}
+
+	if config.MajorTriggerInterval != "" {
+		properties["self-optimizing.major.trigger.interval"] = config.MajorTriggerInterval
+	}
+
+	if config.FullTriggerInterval != "" {
+		properties["self-optimizing.full.trigger.interval"] = config.FullTriggerInterval
+	}
```
hash-data: these are crons property got changed

server/internal/services/optimization/terminal.go
```diff
@@ -0,0 +1,141 @@
...
+		Logs:      logInfo.Logs,
+	}, nil
+}
+
+// pollForCompletion polls the terminal API for SQL execution completion
+func (s *Service) pollForCompletion(ctx context.Context, _, sessionID string) (*dto.LogInfo, error) {
+	const (
+		pollInterval = 1500 * time.Millisecond
+		maxTimeout   = 30 * time.Second
+	)
```
hash-data: let us make constants also it will execute in 30 second right?

server/internal/services/optimization/helpers.go
```diff
@@ -0,0 +1,76 @@
...
+	if resp.Code != 200 {
+		return nil, fmt.Errorf("optimization error (code %d): %s", resp.Code, resp.Message)
+	}
+
+	var result interface{}
+	if err := json.Unmarshal(resp.Result, &result); err != nil {
+		return nil, fmt.Errorf("failed to parse result: %s", err)
+	}
+
+	return result, nil
```
hash-data: let us make sure that logs are different in each function to ease debugging

server/internal/services/common.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (s *AppService) CreateDestinationWithCatalog(ctx context.Context, projectID string, req *dto.CreateDestinationRequest, userID *int) (*DestinationWithCatalogResult, error) {
+	if err := s.etl.CreateDestination(ctx, req, projectID, userID); err != nil {
+		return nil, err
+	}
+
+	result := &DestinationWithCatalogResult{}
+	if s.opt != nil {
+		catalogName, _ := optimization.ExtractCatalogNameFromConfig(req.Config)
```
hash-data: you are unmarshaling two times this one

server/internal/services/common.go
```diff
@@ -0,0 +1,113 @@
...
+		logger.Debugf("Creating catalog[%s] from destination[%s]", catalogName, req.Name)
+		if _, err := s.opt.CreateCatalogFromOLakeConfig(ctx, req.Config, true); err != nil {
+			logger.Errorf("Failed to create catalog[%s]: %s", catalogName, err)
+			result.CatalogErr = err
+		} else {
+			logger.Infof("Catalog[%s] created successfully", catalogName)
+		}
```
hash-data: in case if dest got created and catalog not able to created?

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (s *Service) CreateCatalogFromOLakeConfig(ctx context.Context, configJSON string, byETL bool) (*dto.CatalogResponse, error) {
+	catalogReq, err := CreateOptimizationConfig(configJSON, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to map OLake config to catalog: %s", err)
+	}
+
+	if byETL {
+		catalogReq.Properties["byOLake"] = "true"
```
hash-data: use snake case also rename to olake_created

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,196 @@
...
+	catalog, err := s.GetCatalog(ctx, catalogName)
+	if err != nil {
+		return nil, err
+	}
+
+	return MapoptimizationCatalogToOLakeConfig(catalog)
+}
+
+func (s *Service) CreateCatalogFromOLakeConfig(ctx context.Context, configJSON string, byETL bool) (*dto.CatalogResponse, error) {
+	catalogReq, err := CreateOptimizationConfig(configJSON, false)
```
hash-data: update flag is false in create one as well

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if config.CatalogName == "" {
+		return "", fmt.Errorf("catalog_name not found in config")
+	}
+
+	return config.CatalogName, nil
+}
+
+func CreateOptimizationConfig(configJSON string, update bool) (*dto.CatalogRequest, error) {
```
hash-data:
```suggestion
func UpsertCatalog(configJSON string, update bool) (*dto.CatalogRequest, error) {
```

server/internal/handlers/common.go
```diff
@@ -0,0 +1,150 @@
...
+// @Failure 401 {object} dto.Error401Response "unauthorized"
+// @Failure 500 {object} dto.Error500Response "failed to delete destination"
+// @Router /api/v1/project/{projectid}/destinations/{id} [delete]
+func (h *Handler) DeleteDestinationAndCatalog() {
+	id, err := etl.GetIDFromPath(&h.Controller)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
+		return
+	}
+
```
hash-data: we first remove it from fusion then from olake, as if olake removed and remain in fusion compaction continues

server/internal/models/dto/requests.go
```diff
@@ -133,3 +133,40 @@ type UpdateSyncTelemetryRequest struct {
 type UpdateStateFileRequest struct {
 	StateFile string `json:"state_file" validate:"required"`
 }
+
+type CatalogRequest struct {
+	Name            string            `json:"name"`
+	Type            string            `json:"type"`
+	OptimizerGroup  string            `json:"optimizerGroup,omitempty"`
```
hash-data: in all json keys use snake case

server/internal/models/config.go
```diff
@@ -0,0 +1,74 @@
+package models
```
hash-data, line 1: rename file to catalog_config

server/internal/services/services.go
```diff
@@ -17,12 +23,29 @@ func InitAppService(db *database.Database) (*AppService, error) {
...
+	}
+	enableOptimization := web.AppConfig.DefaultBool(constants.ConfEnableOptimization, false)
+	if enableOptimization {
+		optSvc, err := optimization.InitService()
+		if err != nil {
+			return nil, err
+		}
+
+		appSvc.opt = optSvc
+		appSvc.SyncCatalogs(context.Background())
```
hash-data: can we define context in main and pass? (not required we can add a todo as well)

server/internal/services/optimization/client.go
```diff
@@ -0,0 +1,252 @@
...
+	loginReq.Header.Set("Content-Type", "application/json")
+	loginReq.Header.Set("X-Request-Source", "Web")
+
+	loginResp, err := client.Do(loginReq)
+	if err != nil {
+		return "", "", fmt.Errorf("login failed: %s", err)
+	}
+	defer loginResp.Body.Close()
+
+	cookies := loginResp.Cookies()
```
hash-data, line 220: is it required?

server/conf/app.conf
```diff
@@ -15,4 +15,9 @@ OLAKE_POSTGRES_SSLMODE  = ${OLAKE_POSTGRES_SSLMODE||disable}
...
 sessionon = ${SESSION_ON||true}
 TEMPORAL_ADDRESS = ${TEMPORAL_ADDRESS||temporal:7233}
-CONTAINER_REGISTRY_BASE = ${CONTAINER_REGISTRY_BASE||registry-1.docker.io}
\ No newline at end of file
+CONTAINER_REGISTRY_BASE = ${CONTAINER_REGISTRY_BASE||registry-1.docker.io}
+ENABLE_OPTIMIZATION = ${ENABLE_OPTIMIZATION||false}
+OPTIMIZATION_BASE_URL = ${OPTIMIZATION_BASE_URL||http://127.0.0.1:1630}
+OPTIMIZATION_GROUP = ${OPTIMIZATION_GROUP||spark-container}
+USERNAME = ${USERNAME||admin}
+PASSWORD = ${PASSWORD||password}
```
hash-data: these are same which passed for olake right?

hash-data (changes requested):

server/internal/handlers/optimization/catalog.go
badalprasadsingh: In the code-base, we have `logger.Debugf` across other handler functions of `etl`. Anyways, removed it from optimization handlers.

server/internal/handlers/optimization/catalog.go
badalprasadsingh: In the code-base, we have `logger.Debugf` across other handler functions of `etl`. Anyways, removed it from optimization handlers.

server/internal/services/optimization/client.go
badalprasadsingh: yes, as queryParams are different everytime

for example, in the runs page UI sends pagination values, and since it is not there in the white list of fusion, fusion needs to calculate its signature, `MD5(apiKey + "page1pageSize2" + secret)`

server/internal/services/optimization/catalog.go
badalprasadsingh: yes it will be false, this boolean flag triggers the `setDefaultCatalogProperties` which sets values like "created-at" and "cache-enabled" which should be set only once, during catalog creation.

so if update == false, then only we set these properties.

server/internal/services/optimization/client.go
badalprasadsingh: yes. 

`defer loginResp.Body.Close()`: makes sure the response body should be closed.
`cookies`: the login endpoint creates a session and returns a cookie, without passing the cookies from login to token creation, the request would fail

server/internal/services/common.go
badalprasadsingh: After our discussion of cases where "catalog_name" might not be unique across several destinations, I thought to let destination get created, with the catalog creation fail with the error of asking unique catalog_name. 

But cool, lets make it if created, created for both.

server/internal/services/optimization/catalog.go
badalprasadsingh: why should `CreateOptimizationConfig` be named to `UpsertCatalog`? `CreateOptimizationConfig` job is to create the configuration for Fusion.

Changing its name to `CreateOptConfig`

server/internal/constants/constants.go
```diff
@@ -84,6 +109,13 @@ var (
...
 	// and is updated to "kubernetes" at startup if KUBERNETES_SERVICE_HOST is set.
 	ExecutorEnvironment = "docker"
+
+	// TableFormatList defines supported table formats for catalogs
+	TableFormatList = []string{"ICEBERG"}
+
+	// hard-coding to S3 now, as the other options are "hadoop" & "OSS" for optimization
+	// GCS & ADLS are supported, given the catalog manages the sdk (eg, Lakekeeper with GCS flavour)
+	DefaultStroageType = "S3"
 )
```
hash-data: DefaultOptimizationStorageType

server/internal/constants/constants.go
```diff
@@ -84,6 +109,13 @@ var (
 	// ExecutorEnvironment indicates the runtime environment. Defaults to "docker"
 	// and is updated to "kubernetes" at startup if KUBERNETES_SERVICE_HOST is set.
 	ExecutorEnvironment = "docker"
+
+	// TableFormatList defines supported table formats for catalogs
+	TableFormatList = []string{"ICEBERG"}
```
hash-data: OptimizeTableFormatList

server/internal/constants/constants.go
```diff
@@ -60,6 +65,26 @@ var (
...
+	OptPathTableOptimizingProcesses = "/api/ams/v1/tables/catalogs/%s/dbs/%s/tables/%s/optimizing-processes"
+	OptPathTerminalExecute          = "/api/ams/v1/terminal/catalogs/%s/execute"
+	OptPathTerminalLogs             = "/api/ams/v1/terminal/%s/logs"
+
+	OptMaxTimeout = 30 * time.Second
+	PollInterval  = 1500 * time.Millisecond
+
+	OptMinorCron          = "self-optimizing.minor.trigger.cron"
+	OptMajorCron          = "self-optimizing.major.trigger.cron"
+	OptFullCron           = "self-optimizing.full.trigger.cron"
```
hash-data: let us group all opt constants

server/routes/router.go
```diff
@@ -109,9 +109,29 @@ func Init(h *handlers.Handler) {
...
 	web.Router("/api/v1/platform/releases", etlHandler, "get:GetReleaseUpdates")
+	web.Router("/api/v1/platform/opt/status", h, "get:GetoptimizationStatus")
 
 	// internal routes
 	web.Router("/internal/worker/callback/sync-telemetry", etlHandler, "post:UpdateSyncTelemetry")
 	web.Router("/internal/project/:projectid/jobs/:id/clear-destination/recover", etlHandler, "post:RecoverClearDestination")
 	web.Router("/internal/project/:projectid/jobs/:id/statefile", etlHandler, "put:UpdateStateFile")
+
+	if h.Optimization != nil {
+		optHandler := h.Optimization
```
hash-data, line 120: middleware for these api's ?

server/internal/services/common.go
```diff
@@ -0,0 +1,126 @@
...
+		Writer json.RawMessage `json:"writer"`
+	}
+	if err := json.Unmarshal([]byte(configJSON), &wrapped); err == nil && len(wrapped.Writer) > 0 {
+		return string(wrapped.Writer)
+	}
+	return configJSON
+}
+
+func (s *AppService) SyncCatalogs(ctx context.Context) {
+	destinations, err := s.db.ListDestinations()
```
hash-data: this will run everytime there is restart of UI?

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "invalid config format", err)
+		return
+	}
+
+	result, err := h.opt.CreateCatalog(h.Ctx.Request.Context(), configJSON, false)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, upstreamStatus(err), err.Error(), err)
+		return
+	}
```
hash-data, line 50: is there no other way instead of upstreamStatus?

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// map the catalog details received from opt
+	// to destination config
+	return MapCatalogToDest(catalog)
+}
+
+func (s *Service) GetCatalogInOpt(ctx context.Context, catalogName string) (*dto.CatalogRequest, error) {
```
hash-data: internal function !!

server/internal/services/optimization/helpers.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if err := json.Unmarshal(resp.Result, result); err != nil {
+		return fmt.Errorf("failed to parse result: %s", err)
+	}
+
+	return nil
+}
+
+// (no-result data): performs an HTTP request and validates the response
+func (c *Service) DoAndValidate(ctx context.Context, method, path string, queryParams url.Values, body interface{}) error {
```
hash-data: naming convention ? 
validation is happening in each one

server/internal/handlers/optimization/piggyBacking.go
```diff
@@ -0,0 +1,76 @@
...
+	// Parse upstream response to re-wrap in standard format
+	var upstreamResponse interface{}
+	if err := json.Unmarshal(data, &upstreamResponse); err != nil {
+		utils.ErrorResponse(&h.Controller, http.StatusInternalServerError, "failed to parse upstream response", err)
+		return
+	}
+
+	var optResp dto.OptimizationResponse
+	if jsonErr := json.Unmarshal(data, &optResp); jsonErr == nil && optResp.Code != 0 && optResp.Code != 200 {
+		utils.ErrorResponse(&h.Controller, optResp.Code, optResp.Message, nil)
```
hash-data, line 71: does optResponse exist at this time?

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,117 @@
...
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "catalog name is not present in query params", nil)
+		return "", false
+	}
+	return catalog, true
+}
+
+func (h *Handler) bindJSON(dst interface{}) bool {
+	if err := h.Ctx.BindJSON(dst); err != nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "invalid request body", err)
+		return false
```
hash-data: i would recomment using utils.ErrorReponse in main handler functions

server/internal/services/optimization/mapper.go
```diff
@@ -0,0 +1,206 @@
...
+		req.Properties["created-at"] = time.Now().Format("02 Jan 2006")
```
hash-data: use utc format

server/internal/models/dto/response.go
```diff
@@ -197,3 +199,61 @@ type TelemetryIDResponse struct {
 	TelemetryUserID string `json:"user_id" example:"1234567890abcdef1234567890abcdef"`
 	OlakeUIVersion  string `json:"version" example:"v0.2.5"`
 }
+
+// TODO: convert all optimization struct json-keys to "camel-case" or "snake-case" as required
```
hash-data: not camel-case

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *Service) UpdateCatalog(ctx context.Context, configJSON string) (string, error) {
+	catalogReq, err := s.CreateOptConfig(configJSON, true)
+	if err != nil {
+		return "", fmt.Errorf("failed to create optimization config: %s", err)
+	}
+
+	return catalogReq.Name, s.UpdateCatalogInOpt(ctx, catalogReq.Name, catalogReq)
+}
+
+func (s *Service) UpdateCatalogInOpt(ctx context.Context, catalogName string, req *dto.CatalogRequest) error {
```
hash-data: why new function?

server/internal/services/optimization/client.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func (c *Service) refreshToken() error {
+	apiKey, apiSecret, err := generateToken(c.baseURL, c.username, c.password)
+	if err != nil {
+		return err
+	}
+	c.apiKey = apiKey
+	c.apiSecret = apiSecret
+
```
hash-data: refresh token can be called from multiple api's??

server/internal/services/optimization/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+func validateCatalog(req *dto.CatalogRequest) error {
+	if req.Name == "" {
+		return fmt.Errorf("catalog name is required")
+	}
+	if req.Type == "" {
+		return fmt.Errorf("catalog type is required")
+	}
+
+	if len(req.TableFormatList) == 0 {
+		req.TableFormatList = []string{"ICEBERG"}
```
hash-data: i guess it is constant already created?

server/internal/services/optimization/mapper.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// Set S3 flags - these are always true for S3 storage
+	if config.S3Endpoint != "" {
+		config.S3PathStyle = true
+	}
+
+	return config, nil
+}
+
+func mapoptimizationTypeToOLake(optimizationType string) models.CatalogType {
```
hash-data:
```suggestion
func mapOptimizationTypeToOlakeType(optimizationType string) models.CatalogType {
```

server/internal/services/optimization/mapper.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// mergeMaps returns a new map with base values overridden by src values
+func mergeMaps(base, src map[string]string) map[string]string {
+	result := make(map[string]string, len(base)+len(src))
+	for k, v := range base {
+		result[k] = v
+	}
+	for k, v := range src {
+		result[k] = v
```
hash-data: ?? there is already a map.copy

server/internal/constants/constants.go
```diff
@@ -60,6 +65,26 @@ var (
...
+	// Optimization API paths
+	OptPathCatalogs                 = "/api/ams/v1/catalogs"
+	OptPathCatalogDetail            = "/api/ams/v1/catalogs/%s"
+	OptPathCatalogTables            = "/api/ams/v1/catalogs/%s/databases/%s/tables"
+	OptPathTableDetails             = "/api/ams/v1/tables/catalogs/%s/dbs/%s/tables/%s/details"
+	OptPathTableOptimizingProcesses = "/api/ams/v1/tables/catalogs/%s/dbs/%s/tables/%s/optimizing-processes"
+	OptPathTerminalExecute          = "/api/ams/v1/terminal/catalogs/%s/execute"
+	OptPathTerminalLogs             = "/api/ams/v1/terminal/%s/logs"
+
+	OptMaxTimeout = 30 * time.Second
```
hash-data: OptimizerQueryExecutionTimeout

server/internal/constants/constants.go
```diff
@@ -60,6 +65,26 @@ var (
...
+	PollInterval  = 1500 * time.Millisecond
```
hash-data: OptimizationQueryResultPollTime

server/internal/services/optimization/terminal.go
```diff
@@ -0,0 +1,127 @@
...
+		case <-ticker.C:
+			var logInfo dto.LogInfo
+			if err := s.DoInto(ctx, http.MethodGet, path, url.Values{}, nil, &logInfo); err != nil {
+				return nil, fmt.Errorf("failed to get logs for session %s: %s", sessionID, err)
+			}
+
+			// Check if execution is complete
+			if logInfo.LogStatus == "Finished" || logInfo.LogStatus == "Failed" ||
+				logInfo.LogStatus == "Canceled" || logInfo.LogStatus == "Expired" {
+				return &logInfo, nil
```
hash-data: in case of failed ? we are retuning nil

hash-data (commented):
final review

server/internal/services/common.go
badalprasadsingh: prev: had no other way to run this in a way when optimization was enabled for the first time

removed after "import" feature

server/internal/services/optimization/catalog.go
badalprasadsingh: this function is removed after import destination feature

server/internal/constants/constants.go
badalprasadsingh: keeping it as `OptMaxTimeout` as this same variable is used while passing the Timeout variable in Optimization Service, not only query
```go
func NewClient() (*Service, error) {
	baseURL, username, password, err := getCredentials()
	if err != nil {
		return nil, fmt.Errorf("failed to get credentials: %s", err)
	}

	apiKey, apiSecret, err := generateToken(baseURL, username, password)
	if err != nil {
		return nil, fmt.Errorf("failed to generate token: %s", err)
	}

	return &Service{
		baseURL:   baseURL,
		apiKey:    apiKey,
		apiSecret: apiSecret,
		username:  username,
		password:  password,
		client: &http.Client{
			Timeout: constants.OptMaxTimeout,
		},
	}, nil
}
```

server/internal/services/optimization/terminal.go
badalprasadsingh: discussed

this API behaves like "TestDestination".

server/internal/services/optimization/catalog.go
badalprasadsingh: was required when `etl` used to update the `opt` catalog, 

removed now, after "import" feature

server/internal/handlers/optimization/piggyBacking.go
badalprasadsingh: the `optResp` struct gets filled with data from Fusion's response

server/internal/handlers/optimization/catalog.go
badalprasadsingh: couldn't find a better solution right now, lets do it in V1

server/internal/services/optimization/helpers.go
badalprasadsingh: renamed it to `DoExec`
and common out the validation logic in `parseResponse` function for three of them

server/internal/handlers/optimization/handler.go
```diff
@@ -0,0 +1,39 @@
...
+	h.opt = opt
+}
+
+// returns the HTTP status code carried by an upstream optimzation
+func upstreamStatus(err error) int {
+	var httpErr *services.HTTPError
+	if errors.As(err, &httpErr) {
+		return httpErr.StatusCode
+	}
+	return http.StatusInternalServerError
```
hash-data, line 38: this is not exact status code received from fusion, can you confirm it?

server/internal/handlers/optimization/catalog.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// Convert config to JSON string
+	configJSON, err := utils.MarshalToString(req)
+	if err != nil {
+		utils.ErrorResponse(&h.Controller, badRequestStatusCode, "invalid config format", err)
+		return
+	}
+
+	result, err := h.opt.UpdateCatalog(h.Ctx.Request.Context(), configJSON)
```
hash-data, line 80: get the status code in response only

server/internal/handlers/optimization/piggyBacking.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	}
+
+	utils.RespondJSON(&h.Controller, statusCode, true, "fetched successfully", upstreamResponse)
```
hash-data, line 75: why we are sending success even it can fail in unmarshal?

server/internal/handlers/optimization/catalog.go
hash-data: We will be removing those as well. we need to have cleaner server logs

server/internal/handlers/optimization/piggyBacking.go
```diff
@@ -0,0 +1,76 @@
...
+		body = json.RawMessage(raw)
+	}
+
+	transformedPath := strings.Replace(req.URL.Path, "/api/opt/v1/", "/api/ams/v1/", 1)
+
+	data, statusCode, headers, err := h.opt.ProxyWithHeaders(req.Context(), req.Method, transformedPath, req.URL.Query(), body)
+	if err != nil {
+		if statusCode == 0 {
+			statusCode = http.StatusBadGateway
+		}
```
hash-data: why changing the status code?

server/internal/services/optimization/helpers.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	resp, err := parseResponse(respBody)
+	if err != nil {
+		return nil, err
+	}
+
+	var result interface{}
+	if err := json.Unmarshal(resp.Result, &result); err != nil {
+		return nil, fmt.Errorf("failed to parse result: %s", err)
```
hash-data: at two point same log is there will be dificult to debug let us make sure each log is different and review yourself as well

server/internal/services/optimization/catalog.go
hash-data: prev it was doing both work, update and insert.

server/internal/models/dto/requests.go
hash-data: added a todo for now

server/internal/services/optimization/mapper.go
```diff
@@ -0,0 +1,212 @@
...
+	if _, exists := req.Properties["created-at"]; !exists {
+		req.Properties["created-at"] = time.Now().UTC().Format("02 Jan 2006")
+	}
+}
+
+// mergeMaps returns a new map with base values overridden by src values
+func mergeMaps(base, src map[string]string) map[string]string {
+	result := make(map[string]string, len(base)+len(src))
+	maps.Copy(result, base)
+	maps.Copy(result, src)
```
hash-data: instead of creating a new map we can merge in prev one as well, if its change of value not impacting downstream

server/internal/services/optimization/helpers.go
```diff
@@ -0,0 +1,71 @@
...
+
+	resp, err := parseResponse(respBody)
+	if err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(resp.Result, result); err != nil {
+		return fmt.Errorf("failed to parse result: %s", err)
+	}
+
```
hash-data, line 39: not a correct way at the end things get parsed in interface. Do and DOinto two diff not make sense

hash-data: make it simple one function and result parsing depend on service calling

server/internal/services/optimization/proxy.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+func (s *Service) ProxyWithHeaders(ctx context.Context, method, path string, queryParams url.Values, body json.RawMessage) ([]byte, int, http.Header, error) {
```
hash-data: rename to piggyBacking api, service are for handlers so keep the name same.

server/internal/models/dto/requests.go
```diff
@@ -133,3 +133,40 @@ type UpdateSyncTelemetryRequest struct {
...
+	OptimizeTableFormatList []string          `json:"tableFormatList"`
+	StorageConfig           map[string]string `json:"storageConfig"`
+	AuthConfig              map[string]string `json:"authConfig"`
+	Properties              map[string]string `json:"properties"`
+	TableProperties         map[string]string `json:"tableProperties"`
+}
+
+// LogInfo represents the log information from terminal execution
+type LogInfo struct {
+	LogStatus string   `json:"logStatus"`
```
hash-data: what is log status? give example when it will be true and false

server/internal/services/optimization/terminal.go
```diff
@@ -0,0 +1,122 @@
...
+			if err := s.DoInto(ctx, http.MethodGet, path, url.Values{}, nil, &logInfo); err != nil {
+				return nil, fmt.Errorf("failed to get logs for session %s: %s", sessionID, err)
+			}
+
+			return &logInfo, nil
+		}
+	}
+}
+
+func ConvertMBToBytes(sizeMB int64) string {
```
hash-data: utility

server/go.mod
badalprasadsingh: it's no longer there current PR changes.

I have a slight different version of golang in my laptop, it happens due to `go mod tidy`

server/internal/services/temporal/client.go
badalprasadsingh: not there in the current PR changes

server/internal/services/compaction/services.go
badalprasadsingh: not there in the current PR changes

previously I had separate smaller sub services for optimization service

server/internal/services/optimization/table.go
badalprasadsingh: in the very first section, where tables are listed we show the "last run" of Lite, Medium and Full.

If I search with "all", I need to iterate one by one through all run types to find the "latest run" or "last run" for that type (as after I "full" run might have happened one month back) 

 so I search by type with page=1, and pageSize=1, which gives me just one result of the type (which is the latest type).

server/internal/services/optimization/catalog.go
badalprasadsingh: changed

server/internal/services/optimization/terminal.go
badalprasadsingh: changed

server/internal/models/dto/response.go
badalprasadsingh: removed

server/internal/services/optimization/client.go
badalprasadsingh: lock applied

server/internal/models/dto/requests.go
badalprasadsingh: LogsStatus says if from Fusion's Logs response the status is "Finished", "Failed", etc.

server/internal/handlers/optimization/piggyBacking.go
badalprasadsingh: added the unmarshaling logic

server/internal/services/optimization/mapper.go
badalprasadsingh: simplified it now

server/internal/services/optimization/catalog.go
badalprasadsingh: we are now getting the same status code from fusion

server/internal/handlers/common.go
badalprasadsingh: after import logic, commons are removed

server/internal/handlers/optimization/catalog.go
badalprasadsingh: done

server/internal/handlers/optimization/handler.go
badalprasadsingh: we get exact status code from fusion right now, we need to use `error.As`, in place of `error.Is` as error.Is returns boolean value which is not required as per our use case.

server/internal/handlers/optimization/catalog.go
badalprasadsingh: discussed

server/internal/handlers/optimization/piggyBacking.go
badalprasadsingh: removed

server/internal/services/optimization/helpers.go
badalprasadsingh: Done. Now only DoInto and DoExec exists.

server/internal/handlers/optimization/piggyBacking.go
hash-data: sometime other status code also come on success, are we sure there will be 200 always in success

server/internal/services/optimization/helpers.go
hash-data: still you not read correctly

server/internal/services/optimization/client.go
```diff
@@ -0,0 +1,269 @@
...
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, resp.StatusCode, resp.Header, fmt.Errorf("failed to read response body: %s", err)
+	}
+
+	return respBody, resp.StatusCode, resp.Header, nil
+}
+
+func (s *Service) DoRequest(ctx context.Context, method, path string, queryParams url.Values, body interface{}) ([]byte, error) {
```
hash-data, line 165: why there are warapper on top of it?

server/internal/handlers/optimization/piggyBacking.go
hash-data: what if json error is not nil?

server/internal/handlers/optimization/catalog.go
hash-data: i think you not understand properly

hash-data (approved):

PR #335, opened 2026-03-09
feat: migrate backend from beego to gin

vishalm0509:
# Description

- Replaced Beego HTTP stack with Gin across handlers, routing, middleware, and server startup while keeping the existing API contract and file structure.
- Migrated ORM layer from Beego ORM to GORM, including DB initialization, model/query rewrites, and runtime automigration with existing schema compatibility.

Fixes # (issue)

## Type of change

- [] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


server/internal/constants/constants.go
```diff
@@ -140,15 +114,30 @@ func Init() {
...
 	}
 }
 
-func checkForRequiredVariables(vars []string) {
-	for _, v := range vars {
-		value, err := config.String(v)
-		if err != nil || value == "" {
-			panic("Required config variable not found: ," + v)
+func checkForRequiredVariables() {
+	cfg := appconfig.Load()
```
vikaxsh, line 186: Init() calls appconfig.Load() at line 117 and checkForRequiredVariables() also calls here, pass cfg in the function

server/internal/database/database.go
```diff
@@ -1,99 +1,85 @@
...
-	"github.com/beego/beego/v2/server/web"
-	_ "github.com/beego/beego/v2/server/web/session/postgres" // required for session
-	_ "github.com/lib/pq"                                     // required for registering driver
+	_ "github.com/lib/pq" // required for registering driver
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+	gormlogger "gorm.io/gorm/logger"
 
-	"github.com/datazip-inc/olake-ui/server/internal/constants"
+	"github.com/datazip-inc/olake-ui/server/internal/appconfig"
```
vikaxsh, line 11: how can we avoid  import of appconfig in multiple packages

server/internal/models/dto/validate.go
```diff
@@ -4,28 +4,8 @@ import (
 	"fmt"
 
 	"github.com/datazip-inc/olake-ui/server/internal/constants"
-	"github.com/go-playground/validator/v10"
 )
 
-// ValidateStruct validates any struct that has `validate` tags.
-func Validate(s interface{}) error {
```
vikaxsh, line 11: don't we require this anymore?

server/internal/models/db.go
```diff
@@ -74,42 +80,47 @@ func (d *Destination) TableName() string {
...
 func (c *Catalog) TableName() string {
 	return constants.TableNameMap[constants.CatalogTable]
 }
 
 type Session struct {
-	SessionKey    string    `json:"session_key" orm:"column(session_key);pk;size(64)"`
-	SessionData   string    `json:"session_data" orm:"column(session_data);type(text)"`
-	SessionExpiry time.Time `json:"session_expiry" orm:"column(session_expiry)"`
+	SessionKey    string    `json:"session_key" gorm:"column:session_key;primaryKey;size:64"`
+	SessionData   []byte    `json:"session_data" gorm:"column:session_data;type:bytea"`
```
vikaxsh, line 123: why? Session.SessionData TEXT -> BYTEA

server/internal/models/db.go
```diff
@@ -74,42 +80,47 @@ func (d *Destination) TableName() string {
...
-	StreamsConfig    string       `json:"streams_config" orm:"type(jsonb)"`
-	State            string       `json:"state" orm:"type(jsonb)"`
-	AdvancedSettings *string      `json:"advanced_settings" orm:"column(advanced_settings);type(jsonb);null"`
-	CreatedBy        *User        `json:"created_by" orm:"rel(fk)"`
-	UpdatedBy        *User        `json:"updated_by" orm:"rel(fk)"`
-	ProjectID        string       `json:"project_id" orm:"column(project_id)"`
+	BaseModel
+	ID               int     `json:"id" gorm:"column:id;primaryKey;autoIncrement"`
+	Name             string  `json:"name" gorm:"column:name;size:100"`
+	SourceID         int     `json:"source_id" gorm:"column:source_id"`
```
vikaxsh, line 87: won't it cause any issue changing *Source -> int

server/internal/models/db.go
```diff
@@ -74,42 +80,47 @@ func (d *Destination) TableName() string {
...
-	Frequency        string       `json:"frequency"`
-	StreamsConfig    string       `json:"streams_config" orm:"type(jsonb)"`
-	State            string       `json:"state" orm:"type(jsonb)"`
-	AdvancedSettings *string      `json:"advanced_settings" orm:"column(advanced_settings);type(jsonb);null"`
-	CreatedBy        *User        `json:"created_by" orm:"rel(fk)"`
-	UpdatedBy        *User        `json:"updated_by" orm:"rel(fk)"`
-	ProjectID        string       `json:"project_id" orm:"column(project_id)"`
+	BaseModel
+	ID               int     `json:"id" gorm:"column:id;primaryKey;autoIncrement"`
+	Name             string  `json:"name" gorm:"column:name;size:100"`
```
vikaxsh, line 86: beego orm by default creates Stirng as varchar(255) in db, would it be same i goorm?
https://beegodoc.com/en-US/developing/orm/model.html#advance-usage

server/internal/services/etl/auth.go
```diff
@@ -16,13 +17,13 @@ func (s *ETLService) Login(ctx context.Context, username, password string) (*mod
 	user, err := s.db.GetUserByUsername(username)
 	if err != nil {
 		if strings.Contains(err.Error(), "no row found") {
-			return nil, fmt.Errorf("user not found: %s", err)
+			return nil, fmt.Errorf("%w: %v", constants.ErrUserNotFound, err)
```
vikaxsh: Can we use a uniform error formatting across files, such as using %w instead of %s

server/internal/services/etl/job.go
```diff
@@ -179,28 +182,16 @@ func (s *ETLService) UpdateJob(ctx context.Context, req *dto.UpdateJobRequest, p
...
-	// Start transaction
-	tx, err := s.db.BeginTx()
-	if err != nil {
-		return fmt.Errorf("failed to begin transaction: %s", err)
-	}
-	defer func() {
-		if err := tx.RollbackUnlessCommit(); err != nil {
-			logger.Errorf("failed to rollback transaction for job[%d]: %s", existingJob.ID, err)
-		}
-	}()
```
vikaxsh, line 191: Why was this removed?

server/internal/services/etl/job.go
```diff
@@ -210,30 +201,24 @@ func (s *ETLService) UpdateJob(ctx context.Context, req *dto.UpdateJobRequest, p
...
-
 	if req.AdvancedSettings != nil {
 		b, err := json.Marshal(req.AdvancedSettings)
 		if err != nil {
 			return fmt.Errorf("failed to serialise advanced_settings: %s", err)
 		}
 		updateParams["advanced_settings"] = string(b)
+	} else {
+		updateParams["advanced_settings"] = nil
 	}
```
vikaxsh, line 219: why?

server/routes/router.go
```diff
@@ -0,0 +1,70 @@
+package routes
```
vikaxsh, line 1: why have we changed file name?

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,130 @@
...
+		defer cancel()
+		return s.httpServer.Shutdown(shutdownCtx)
+	case err := <-errCh:
+		return err
+	}
+}
+
+func configureMode(cfg *appconfig.Config) {
+	switch cfg.RunMode {
+	case "dev", "localdev":
```
vikaxsh: since by default run mode is dev, execution will always enter this block

server/internal/appconfig/appconfig.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/spf13/viper"
+)
+
+type Config struct {
+	AppName               string
+	HTTPPort              string
+	RunMode               string
+	CopyRequestBody       bool
+	MaxMemory             int64
+	MaxUploadSize         int64
```
vikaxsh: aren't we using these anymore?

server/internal/database/database.go
```diff
@@ -1,99 +1,85 @@
...
-	if web.BConfig.WebConfig.Session.SessionOn {
-		web.BConfig.WebConfig.Session.SessionName = "olake-session"
-		web.BConfig.WebConfig.Session.SessionProvider = "postgresql"
-		web.BConfig.WebConfig.Session.SessionProviderConfig = uri
-		web.BConfig.WebConfig.Session.SessionCookieLifeTime = 30 * 24 * 60 * 60 // 30 days
+		return nil, fmt.Errorf("failed to connect to postgres: %s", err)
 	}
 
-	// register session user
-	gob.Register(constants.SessionUserID)
```
vikaxsh, line 49: beego version was storing session data gob-encoded bytes (BYTEA), what would happen now?

server/internal/database/destination.go
```diff
@@ -92,16 +102,18 @@ func (db *Database) UpdateDestination(destination *models.Destination) error {
 		return fmt.Errorf("failed to encrypt destination[%d] config: %s", destination.ID, err)
 	}
 	destination.Config = eConfig
-	_, err = db.ormer.Update(destination)
-	return err
+	return db.conn.Updates(destination).Error
```
vikaxsh: would it perform full column update?

server/internal/models/db.go
```diff
@@ -8,18 +8,18 @@ import (
 
 // BaseModel with common fields
 type BaseModel struct {
-	CreatedAt time.Time  `json:"created_at" orm:"column(created_at);auto_now_add;type(datetime)"`
-	UpdatedAt time.Time  `json:"updated_at" orm:"column(updated_at);auto_now;type(datetime)"`
-	DeletedAt *time.Time `json:"deleted_at,omitempty" orm:"column(deleted_at);null;type(datetime)"`
+	CreatedAt time.Time  `json:"created_at" gorm:"column:created_at;autoCreateTime"`
+	UpdatedAt time.Time  `json:"updated_at" gorm:"column:updated_at;autoUpdateTime"`
+	DeletedAt *time.Time `json:"deleted_at,omitempty" gorm:"column:deleted_at"`
```
vikaxsh, line 13: Shouldn’t we use gorm.DeletedAt here? Otherwise it’ll perform a hard delete in the DB, and methods like Find, First, etc. will stop filtering out previously deleted records (like Beego used to)

server/internal/database/job.go
```diff
@@ -214,10 +208,16 @@ func (db *Database) IsNameUniqueInProject(ctx context.Context, projectID, name s
 		return false, fmt.Errorf("invalid table type: %v", tableType)
 	}
 
-	count, err := db.ormer.QueryTableWithCtx(ctx, tableName).
-		Filter("name", name).
-		Filter("project_id", projectID).
-		Count()
+	countQuery := db.conn.Table(tableName)
+	if ctx != nil {
```
vikaxsh: when are we expecting nil ctx?

server/internal/database/project.go
```diff
@@ -34,23 +33,26 @@ func (db *Database) UpsertProjectSettingsModel(settings *models.ProjectSettings)
...
-	if err == orm.ErrNoRows {
+	existing := &models.ProjectSettings{}
+	err := db.conn.Where("project_id = ?", settings.ProjectID).First(existing).Error
+	if err != nil && err != gorm.ErrRecordNotFound {
+		return fmt.Errorf("failed to lookup project settings project_id[%s]: %s", settings.ProjectID, err)
+	}
+	if err == gorm.ErrRecordNotFound {
 		settings.ID = 0
-		if _, err := db.ormer.Insert(settings); err != nil {
+		if err := db.conn.Create(settings).Error; err != nil {
```
vikaxsh: can we eliminate 2 db calls here and use clause.OnConflict for atomic upsert and prevent race condition as well

server/internal/database/project.go
```diff
@@ -34,23 +33,26 @@ func (db *Database) UpsertProjectSettingsModel(settings *models.ProjectSettings)
...
-	}
 
-	// Record exists, update it
 	settings.ID = existing.ID
-	if _, err := db.ormer.Update(settings); err != nil {
+	if err := db.conn.Model(&models.ProjectSettings{}).
+		Where("id = ?", settings.ID).
+		Updates(map[string]any{
+			"project_id":        settings.ProjectID,
+			"webhook_alert_url": settings.WebhookAlertURL,
```
vikaxsh: why do we need a hardcoded map here? can we think of something else here?

server/internal/database/project.go
```diff
@@ -34,23 +33,26 @@ func (db *Database) UpsertProjectSettingsModel(settings *models.ProjectSettings)
...
-		return fmt.Errorf("failed to lookup project settings project_id[%s]: %s", settings.ProjectID, err)
-	}
 
-	// Record exists, update it
 	settings.ID = existing.ID
-	if _, err := db.ormer.Update(settings); err != nil {
+	if err := db.conn.Model(&models.ProjectSettings{}).
+		Where("id = ?", settings.ID).
+		Updates(map[string]any{
+			"project_id":        settings.ProjectID,
```
vikaxsh: why do we need to update project id?

server/internal/database/user.go
```diff
@@ -3,15 +3,16 @@ package database
...
 	"golang.org/x/crypto/bcrypt"
 
-	"github.com/datazip-inc/olake-ui/server/internal/constants"
 	"github.com/datazip-inc/olake-ui/server/internal/models"
 )
 
 func (db *Database) GetUserByUsername(username string) (*models.User, error) {
 	var user models.User
-	err := db.ormer.QueryTable(constants.TableNameMap[constants.UserTable]).Filter("username", username).One(&user)
+	err := db.conn.Where("username = ?", username).First(&user).Error
```
vikaxsh, line 17: add a error nil check here

server/internal/database/destination.go
```diff
@@ -29,13 +30,16 @@ func (db *Database) CreateDestination(destination *models.Destination) error {
...
+	return db.conn.Create(destination).Error
 }
 
 func (db *Database) ListDestinations() ([]*models.Destination, error) {
 	var destinations []*models.Destination
-	_, err := db.ormer.QueryTable(constants.TableNameMap[constants.DestinationTable]).RelatedSel().OrderBy(constants.OrderByUpdatedAtDesc).All(&destinations)
+	err := db.conn.
+		Preload("CreatedBy").
+		Preload("UpdatedBy").
+		Order("updated_at DESC").
```
vikaxsh, line 42: Would it be better to create constants for all such fields used across the queries?

server/internal/database/database.go
```diff
@@ -1,99 +1,85 @@
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to build postgres uri: %s", err)
 	}
 
-	err = orm.RegisterDriver("postgres", orm.DRPostgres)
-	if err != nil {
-		return nil, fmt.Errorf("failed to register postgres driver: %s", err)
+	logLevel := gormlogger.Warn
+	if cfg.RunMode == "dev" || cfg.RunMode == "localdev" {
+		logLevel = gormlogger.Info
```
vikaxsh, line 30: why?

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,130 @@
...
+const frontendDistPath = "/opt/frontend/dist"
+
+type Server struct {
+	engine     *gin.Engine
+	httpServer *http.Server
+}
+
+func New(cfg *appconfig.Config, h *handlers.Handler) *Server {
+	engine := gin.New()
+	engine.Use(gin.Recovery())
```
vikaxsh: Shouldn't we also add gin.Logger() to log incoming API requests?

server/main.go
```diff
@@ -62,14 +66,22 @@ func main() {
...
 	}
 
-	routes.Init(handlers.NewHandler(appSvc))
-	if key, _ := web.AppConfig.String(constants.ConfEncryptionKey); key == "" {
+	cfg := appconfig.Load()
+	if cfg.EncryptionKey == "" {
 		logger.Warn("Encryption key is not set. This is not recommended for production environments.")
 	}
 
-	if web.BConfig.RunMode == "dev" || web.BConfig.RunMode == "staging" {
```
vikaxsh, line 70: "staging" is not included anywhere, can we check isn't it require anymore

server/internal/handlers/user.go
```diff
@@ -71,28 +69,26 @@ func (h *Handler) GetAllUsers() {
 // @Failure 401 {object} dto.Error401Response "unauthorized"
 // @Failure 500 {object} dto.Error500Response "failed to update user"
 // @Router /api/v1/users/{id} [put]
-func (h *Handler) UpdateUser() {
-	id, err := GetIDFromPath(&h.Controller)
+func (h *Handler) UpdateUser(c *gin.Context) {
+	id, err := strconv.Atoi(c.Param("id"))
```
vikaxsh: use getIDParam

server/internal/handlers/user.go
```diff
@@ -19,26 +20,25 @@ import (
...
+func (h *Handler) CreateUser(c *gin.Context) {
 	var req models.User
-	if err := UnmarshalAndValidate(h.Ctx.Input.RequestBody, &req); err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
+	if err := bindAndValidate(c, &req); err != nil {
+		errorResponse(c, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
 		return
 	}
-
 	if req.Username == "" || req.Email == "" || req.Password == "" {
```
vikaxsh: let's remove Manual Validation

server/internal/handlers/user.go
```diff
@@ -19,26 +20,25 @@ import (
...
 
-	if err := h.etl.CreateUser(h.Ctx.Request.Context(), &req); err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusInternalServerError, fmt.Sprintf("failed to create user: %s", err), err)
+	if err := h.etl.CreateUser(c.Request.Context(), &req); err != nil {
+		errorResponse(c, http.StatusInternalServerError, fmt.Sprintf("failed to create user: %s", err), err)
 		return
 	}
 
-	utils.SuccessResponse(&h.Controller, "user created successfully", req)
+	successResponse(c, "user created successfully", req)
```
vikaxsh: can we avoid sending password back

server/internal/handlers/etl/swagger.go
```diff
@@ -1,23 +1,13 @@
...
 
 // ServeSwagger serves the Swagger UI and static swagger files
-func (h *Handler) ServeSwagger() {
-	url := h.Ctx.Input.URL()
-
-	// Redirect /swagger to /swagger/ for http-swagger compatibility
-	if url == "/swagger" {
-		h.Redirect("/swagger/", http.StatusMovedPermanently)
-		return
-	}
```
vikaxsh, line 17: don't we require anymore

server/internal/handlers/etl/source.go
```diff
@@ -183,25 +164,19 @@ func (h *Handler) UpdateSource() {
...
 		return
 	}
-
 	logger.Debugf("Delete source initiated source_id[%d]", id)
-
-	resp, err := h.etl.DeleteSource(h.Ctx.Request.Context(), id)
+	resp, err := h.etl.DeleteSource(c.Request.Context(), id)
 	if err != nil {
-		if errors.Is(err, constants.ErrSourceNotFound) {
-			utils.ErrorResponse(&h.Controller, http.StatusNotFound, fmt.Sprintf("source not found: %s", err), err)
```
vikaxsh, line 198: we might require this as need to send correct response code

server/internal/handlers/etl/source.go
```diff
@@ -283,25 +250,19 @@ func (h *Handler) GetSourceCatalog() {
...
 // @Router /api/v1/project/{projectid}/sources/versions [get]
-func (h *Handler) GetSourceVersions() {
-	projectID, err := GetProjectIDFromPath(&h.Controller)
-	if err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
-		return
-	}
-
-	sourceType := h.GetString("type")
-	logger.Debugf("Get source versions initiated project_id[%s] source_type[%s]", projectID, sourceType)
```
vikaxsh, line 294: add this back, check other places as well, if removed

server/internal/handlers/session.go
```diff
@@ -0,0 +1,115 @@
...
+		return 0, false
+	}
+
+	sessionID, err := c.Cookie(sessionCookieName)
+	if err != nil || sessionID == "" {
+		return 0, false
+	}
+
+	var rawPayload []byte
+	err = s.db.QueryRow(`
```
vikaxsh: why are we firing db queries from handlers?

server/internal/handlers/session.go
```diff
@@ -0,0 +1,115 @@
...
+	if !store.enabled {
+		return store, nil
+	}
+
+	dsn, err := database.BuildPostgresURIFromConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build postgres URI for session store: %w", err)
+	}
+
+	dbConn, err := sql.Open("postgres", dsn)
```
vikaxsh: why do we need separate db connection for session

server/internal/handlers/platform.go
```diff
@@ -18,21 +18,18 @@ import (
 // @Failure 401 {object} dto.Error401Response "unauthorized"
```
vikaxsh: where are we returning these response code in this api? please once verify for other apis as well

server/internal/handlers/etl/auth.go
```diff
@@ -23,77 +22,32 @@ import (
...
-
 	logger.Debugf("Login initiated username[%s]", req.Username)
 
-	user, err := h.etl.Login(h.Ctx.Request.Context(), req.Username, req.Password)
+	user, err := h.etl.Login(c.Request.Context(), req.Username, req.Password)
 	if err != nil {
-		switch {
-		case errors.Is(err, constants.ErrUserNotFound):
-			utils.ErrorResponse(&h.Controller, http.StatusUnauthorized, fmt.Sprintf("user not found, sign up first: %s", err), err)
-		case errors.Is(err, constants.ErrInvalidCredentials):
```
vikaxsh, line 40: don't we require further?

server/internal/handlers/etl/auth.go
```diff
@@ -23,77 +22,32 @@ import (
...
-			utils.ErrorResponse(&h.Controller, http.StatusUnauthorized, fmt.Sprintf("Invalid session: %s", err), err)
-			return
-		}
-	}
-
-	utils.SuccessResponse(&h.Controller, "authenticated successfully", nil)
+	successResponse(c, "login successful", dto.LoginResponse{Username: user.Username})
 }
 
-// func (h *Handler) Logout() {
```
vikaxsh, line 84: why?

server/internal/handlers/auth.go
```diff
@@ -103,44 +57,78 @@ func (h *Handler) CheckAuth() {
 // @Failure 409 {object} dto.Error409Response "user already exists"
 // @Failure 500 {object} dto.Error500Response "failed to create user"
 // @Router /signup [post]
-func (h *Handler) Signup() {
+func (h *Handler) Signup(c *gin.Context) {
 	var req models.User
```
vikaxsh: can we create dto's for these as well?

server/internal/handlers/destination.go
```diff
@@ -18,19 +19,19 @@ import (
...
-	items, err := h.etl.ListDestinations(h.Ctx.Request.Context(), projectID)
+	logger.Debugf("Get all destinations initiated project_id[%s]", projectID)
+	items, err := h.etl.ListDestinations(c.Request.Context(), projectID)
 	if err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusInternalServerError, fmt.Sprintf("failed to get destinations: %s", err), err)
+		errorResponse(c, http.StatusInternalServerError, fmt.Sprintf("failed to get destinations: %s", err), err)
 		return
 	}
-	utils.SuccessResponse(&h.Controller, "Destinations listed successfully", items)
+	successResponse(c, "Destinations listed successfully", items)
```
vikaxsh:
```suggestion
	successResponse(c, "destinations listed successfully", items)
```

server/internal/handlers/etl/job.go
```diff
@@ -141,59 +124,33 @@ func (h *Handler) CreateJob() {
...
 	var req dto.UpdateJobRequest
-	if err := UnmarshalAndValidate(h.Ctx.Input.RequestBody, &req); err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
+	if err := bindAndValidate(c, &req); err != nil {
+		errorResponse(c, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
 		return
 	}
-
-	if req.Source.ID == nil {
-		if err := dto.ValidateSourceType(req.Source.Type); err != nil {
```
vikaxsh, line 170: don't we require source type validation anymore?

server/internal/handlers/etl/job.go
```diff
@@ -79,55 +77,40 @@ func (h *Handler) GetJob() {
...
 	}
-
-	// Conditional validation
-	if req.Source.ID == nil {
+	if req.Source != nil && req.Source.ID == nil {
 		if err := dto.ValidateSourceType(req.Source.Type); err != nil {
-			utils.ErrorResponse(&h.Controller, http.StatusBadRequest, fmt.Sprintf("failed to validate request: %s", err), err)
-			return
-		}
-		if req.Source.Name == "" || req.Source.Version == "" || req.Source.Config == "" {
```
vikaxsh, line 107: don't we require this check anymore?

server/internal/handlers/etl/job.go
```diff
@@ -580,73 +561,23 @@ func (h *Handler) UpdateSyncTelemetry() {
...
-
-	h.Ctx.Output.Header("Content-Type", "application/gzip")
-	h.Ctx.Output.Header("Content-Disposition", fmt.Sprintf("attachment; filename=%q", filename))
-	h.Ctx.Output.Header("Cache-Control", "no-cache")
-	h.Ctx.Output.Header("X-Content-Type-Options", "nosniff")
-	// Expose Content-Disposition header so browser JS can access filename for download
-	h.Ctx.Output.Header("Access-Control-Expose-Headers", "Content-Disposition")
-
-	if err := h.etl.StreamLogArchive(id, filePath, h.Ctx.ResponseWriter); err != nil {
-		logger.Errorf("failed to stream log archive job_id[%d]: %s", id, err)
```
vikaxsh, line 645: add this back

server/internal/constants/constants.go
vishalm0509: Config is loaded once at startup and then every time we call `appconfig.Load()` it returns the already loaded config. Therefore, no re-computation of all the envs.

```go
// this loads first on startup
var cfg = loadConfig()

// this serves the already loaded cfg
func Load() Config {
	return cfg
}
```

server/internal/database/database.go
vishalm0509: Config is loaded once at startup and then every time we call `appconfig.Load()` it returns the already loaded config. Therefore, no re-computation of all the envs.

```go
// this loads first on startup
var cfg = loadConfig()

// this serves the already loaded cfg
func Load() Config {
	return cfg
}
```

--- 

It is similar to that of `web.AppConfig.String(constants.ConfOLakePostgresUser)` we import `web` everywhere. `appconfig` is similar to that

server/internal/models/dto/validate.go
vishalm0509: No, here `c.ShouldBindJSON()` handles bindings.

```go
func bindAndValidate(c *gin.Context, target interface{}) error {
	return c.ShouldBindJSON(target)
}
```

We have replaced `Username string json:"username" validate:"required" example:"admin"` (**validate**) with this (**binding**) which does the same work.

server/internal/database/database.go
vikaxsh: can we think of any alternate to avoid this now

server/internal/models/db.go
vishalm0509: In the earlier code beego also wrote data to postgres in `[]byte` 

```go
	// Add session table if sessions are enabled
	if web.BConfig.WebConfig.Session.SessionOn {
		_, err = orm.NewOrm().Raw(`CREATE TABLE IF NOT EXISTS session (
    session_key VARCHAR(64) PRIMARY KEY,
    session_data BYTEA,
    session_expiry TIMESTAMP WITH TIME ZONE
);`).Exec()
```

Screenshot 2026-03-12 at 12 57 21

server/internal/models/db.go
vishalm0509: This is **FK** reference ID, we still can access the Source object with `preLoad`

```go
	Source      *Source      `json:"source,omitempty" gorm:"foreignKey:SourceID;references:ID"`
	Destination *Destination `json:"destination,omitempty" gorm:"foreignKey:DestID;references:ID"`
	CreatedBy   *User        `json:"created_by,omitempty" gorm:"foreignKey:CreatedByID;references:ID"`
	UpdatedBy   *User        `json:"updated_by,omitempty" gorm:"foreignKey:UpdatedByID;references:ID"`
```

server/internal/models/db.go
vishalm0509: Here, we are specifying `size:100` so it would be `VARCHAR(100`, similarly `size:N` becomes `VARCHAR(N)`

```
Name             string  `json:"name" gorm:"column:name;size:100"`
```

if we don't specify size it would be mapped to `TEXT` Datatype.

image

server/internal/services/etl/auth.go
vishalm0509: Since we are using `errors.Is()` in the Login Handler we need `%w` so that `errors.Is()` can detect it.

and for `%v` we can standardize to either `%v` or `%s`

server/internal/services/etl/job.go
vishalm0509: Because transaction make any difference in this use-case as it was being used only by `s.db.updateJob()` and it behaves the same way if remove as well. 

added a **TODO** to properly handle this case.

server/internal/services/etl/job.go
vishalm0509: If we don't set it to `nil` on `req.AdvancedSettings == nil` then the previous data will be persisted in the database. Now if user removes AdvancedSettings and save it will be set to `NULL`

server/routes/router.go
vishalm0509: Reverted back to `router.go`

server/internal/httpserver/server.go
vishalm0509: Right, now Debug mode runs only on `RunMode="localdev"`

server/internal/appconfig/appconfig.go
vishalm0509: **CopyRequestBody**

`CopyRequestBody` is not needed in Gin, if we need to re-read a request body we should use `ShouldBindBodyWith()`

https://stackoverflow.com/questions/62736851/go-gin-read-request-body-many-times#:~:text=According%20to%20the%20ShouldBindBodyWith%20documentation%2C%20it%20only,them%20inevitably%20leads%20to%20an%20EOF%20error.

---

Added the config to Gin for  `MaxMemory` and `MaxUploadSize`

server/internal/database/database.go
vishalm0509: The old session will be invalid and requires a re-auth. And we are also storing the session data in `BYTEA` also the cookie name is new now `olake-session`

server/internal/database/destination.go
vishalm0509: Yes

server/internal/models/db.go
vishalm0509: Beego also did hard delete there was no soft-delete earlier. It's the same behaviour

server/internal/database/job.go
vishalm0509: None :)

server/internal/database/project.go
vishalm0509: Right

server/internal/database/project.go
vishalm0509: Made changes with the above comment

server/internal/database/user.go
vishalm0509: We are returning the error right, so if it's `nil` it will be handled in the service layer

server/internal/database/database.go
vishalm0509: - gormlogger.Info is noisy (every SQL statement) is logged.
- Keeping dev at Warn reduces log volume and avoids leaking query details in Release env.

server/internal/httpserver/server.go
vishalm0509: Done

server/main.go
vishalm0509: Not needed for now. All the RunMode (`dev` | `staging` | `prdd`) are handled in the default case of switch statements and defaults to ReleaseMode

server/internal/handlers/user.go
vishalm0509: Done

server/internal/handlers/user.go
vishalm0509: We can create a separate DTO for this then ?

server/internal/database/destination.go
vishalm0509: I have seen this pattern in other tools as well.

server/internal/handlers/user.go
vishalm0509: Done, created a dto

server/internal/handlers/user.go
vishalm0509: created

server/internal/handlers/etl/swagger.go
vishalm0509: No

server/internal/handlers/etl/source.go
vishalm0509: Yes, added back

server/internal/handlers/etl/auth.go
vishalm0509: It was commented earlier right

server/internal/handlers/etl/job.go
vishalm0509: NEeded

server/internal/handlers/etl/job.go
vishalm0509: Done

server/routes/router.go
```diff
@@ -1,115 +1,70 @@
...
-	web.Router("/api/v1/project/:projectid/jobs/:id/stream-difference", h, "post:GetStreamDifference")
-
-	// Project settings routes
-	web.Router("/api/v1/project/:projectid/settings", h, "put:UpsertProjectSettings")
-	web.Router("/api/v1/project/:projectid/settings", h, "get:GetProjectSettings")
-
-	// validation routes
-	web.Router("/api/v1/project/:projectid/check-unique", h, "post:CheckUniqueName")
-
-	// platform routes
```
vikaxsh, line 110: add these comments back

server/internal/appconfig/appconfig.go
vikaxsh: can we remove CopyRequestBody

server/internal/handlers/utils.go
```diff
@@ -1,49 +1,72 @@
...
-	id, err := strconv.Atoi(idStr)
+func getProjectID(c *gin.Context) (string, error) {
+	projectID := c.Param("projectid")
+	if projectID == "" {
+		return "", fmt.Errorf("project id is required")
+	}
+	return projectID, nil
+}
+
+func getIDParam(c *gin.Context, key string) (int, error) {
```
vikaxsh: key always will be "id" here, can we avoid passing in function

server/internal/database/database.go
vishalm0509: We have decided to use this now

server/internal/models/db.go
vishalm0509: Done made it to size(100) and size(255) wherever necessary

server/main.go
vishalm0509: it will be handled in the `default` case

server/internal/handlers/etl/auth.go
vishalm0509: returning errors.Is() now from database level

server/internal/handlers/etl/auth.go
vishalm0509: Added back

server/internal/database/job.go
```diff
@@ -61,14 +65,19 @@ func (db *Database) decryptJobSliceConfig(jobs []*models.Job) error {
 
 // Create a new job
 func (db *Database) CreateJob(job *models.Job) error {
-	_, err := db.ormer.Insert(job)
-	return err
+	return db.conn.Create(job).Error
 }
 
 // GetAll retrieves all jobs
 func (db *Database) ListJobs() ([]*models.Job, error) {
```
vikaxsh: dead code

server/internal/database/user.go
```diff
@@ -11,43 +14,63 @@ import (
...
+		return err
+	}
+	return nil
 }
 
 func (db *Database) ListUsers() ([]*models.User, error) {
 	var users []*models.User
-	_, err := db.ormer.QueryTable(constants.TableNameMap[constants.UserTable]).All(&users)
+	err := db.conn.Find(&users).Error
 	return users, err
```
vikaxsh, line 48: fmt.Errorf("failed to find user: %s", err)

server/internal/handlers/etl/job.go
```diff
@@ -517,59 +504,116 @@ func (h *Handler) GetJobTasks() {
...
+		return
+	}
+
+	c.Header("Content-Type", "application/gzip")
+	c.Header("Content-Disposition", fmt.Sprintf("attachment; filename=%q", filename))
+	c.Header("Cache-Control", "no-cache")
+	c.Header("X-Content-Type-Options", "nosniff")
+	c.Header("Access-Control-Expose-Headers", "Content-Disposition")
+	c.Status(http.StatusOK)
+	if err := h.etl.StreamLogArchive(id, filePath, c.Writer); err != nil {
```
vikaxsh: what if StreamLogArchive returns an error without writing any bytes, it would return 200 OK with an empty response

server/internal/handlers/etl/source.go
```diff
@@ -212,29 +199,26 @@ func (h *Handler) DeleteSource() {
...
 		return
 	}
-
 	logger.Infof("Test source connection initiated source_type[%s] source_version[%s]", req.Type, req.Version)
-
-	result, logs, err := h.etl.TestSourceConnection(h.Ctx.Request.Context(), &req)
+	result, logs, err := h.etl.TestSourceConnection(c.Request.Context(), &req)
 	if err != nil {
-		utils.ErrorResponse(&h.Controller, http.StatusInternalServerError, fmt.Sprintf("failed to verify credentials: %s", err), err)
+		utils.ErrorResponse(c, http.StatusInternalServerError, fmt.Sprintf("failed to verify credentials: %s", err), err)
```
vikaxsh, line 219: return 400 here

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,178 @@
...
+		IdleTimeout:       120 * time.Second,
+	}
+
+	return &Server{
+		engine:     engine,
+		httpServer: server,
+	}
+}
+
+func (s *Server) Engine() *gin.Engine {
```
vikaxsh: where are we using this?

server/internal/services/etl/user.go
```diff
@@ -2,15 +2,27 @@ package etl
...
 
+	"github.com/datazip-inc/olake-ui/server/internal/constants"
 	"github.com/datazip-inc/olake-ui/server/internal/models"
+	"golang.org/x/crypto/bcrypt"
 )
 
 // User-related methods on AppService
 
-func (s Service) CreateUser(_ context.Context, req *models.User) error {
+func (s *Service) CreateUser(_ context.Context, req *models.User) error {
```
vikaxsh: why pointer receiver?

server/routes/router.go
```diff
@@ -1,137 +1,128 @@
...
-	web.Router("/api/v1/project/:projectid/settings", etlHandler, "put:UpsertProjectSettings")
-	web.Router("/api/v1/project/:projectid/settings", etlHandler, "get:GetProjectSettings")
-
-	// validation routes
-	web.Router("/api/v1/project/:projectid/check-unique", etlHandler, "post:CheckUniqueName")
-
-	// platform routes
-	web.Router("/api/v1/platform/releases", etlHandler, "get:GetReleaseUpdates")
-	web.Router("/api/v1/platform/opt/status", h, "get:GetoptimizationStatus")
+	etl := engine.Group("/api/v1")
```
vikaxsh: in beego we were applying auth for /api/* path

server/routes/router.go
vishalm0509: Done

server/internal/handlers/utils.go
vishalm0509: Right, but for other modules will it still be `id` always ? since we are using this as a global util for all the modules

vishalm0509: I have made changes anyway since the name is explicitly `GetIDParam` but the above is something we should think about

server/internal/database/job.go
vishalm0509: Yes, removed.

server/internal/database/user.go
vishalm0509: here, err can be nil as well. This is the design we are following in the database layer. 

We are adding a proper error message in the service layer to which this is appended

server/internal/handlers/etl/job.go
vishalm0509: Yes, made changes

server/internal/httpserver/server.go
vishalm0509: removed, was using it earlier

server/internal/services/etl/user.go
vishalm0509: made changes

server/routes/router.go
vishalm0509:
```go
	// Apply auth middleware to protected routes
	web.InsertFilter("/api/*", web.BeforeRouter, middleware.AuthMiddleware)
```

Auth middlware only applied to routes with prefix `/api`

server/internal/database/source.go
```diff
@@ -93,14 +104,18 @@ func (db *Database) UpdateSource(source *models.Source) error {
 		return fmt.Errorf("failed to encrypt source config id[%d]: %s", source.ID, err)
 	}
 	source.Config = eConfig
-	_, err = db.ormer.Update(source)
-	return err
+	return db.conn.Updates(source).Error
```
vikaxsh: gorm.Updates() Silently Skips Zero-Values on Structs
please check following doc
https://gorm.io/docs/update.html#Updates-multiple-columns

server/internal/appconfig/appconfig.go
```diff
@@ -0,0 +1,78 @@
...
+
+func Load() Config {
+	return cfg
+}
+
+func loadConfig() Config {
+	v := viper.New()
+
+	// Note: config priority: env variables -> file (app.yaml)
+	v.SetConfigFile("./conf/app.yaml")
```
vikaxsh: can it cause any issue in k8s where all configs come from env vars

server/internal/handlers/session.go
```diff
@@ -0,0 +1,83 @@
...
+
+	sessionID := utils.ULID()
+	expiresAt := time.Now().Add(sessionMaxAgeDays * 24 * time.Hour)
+
+	payload, err := json.Marshal(sessionPayload{UserID: userID})
+	if err != nil {
+		return fmt.Errorf("failed to marshal session payload: %w", err)
+	}
+
+	err = s.db.UpsertSession(sessionID, payload, expiresAt)
```
vikaxsh: why upsert not create

server/routes/router.go
vikaxsh: yes
now it is applying to -> etl := engine.Group("/api/v1"),etl.Use(h.AuthMiddleware())

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,174 @@
...
+		return s.httpServer.Shutdown(shutdownCtx)
+	case err := <-errCh:
+		return err
+	}
+}
+
+func configureMode(cfg *appconfig.Config) {
+	switch cfg.RunMode {
+	case "localdev":
+		gin.SetMode(gin.DebugMode)
```
vikaxsh, line 84: gin.SetMode() is a global setter that must be called before any gin.New() or gin.Default() call. Calling it after the engine is already created and middleware is already registered changes the global mode but has no effect on the already-constructed engine.

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func configureStaticFrontend(engine *gin.Engine) {
+	engine.Static("/assets", filepath.Join(frontendDistPath, "assets"))
+	engine.StaticFile("/favicon.ico", filepath.Join(frontendDistPath, "favicon.ico"))
+}
+
+func configureNoRoute(engine *gin.Engine, cfg *appconfig.Config, h *handlers.Handler) {
+	moduleHandlers := make([]routes.ModuleNoRouteHandler, 0, 1)
+	if h != nil && h.Optimization != nil {
```
vikaxsh: h would be always non nil here

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,174 @@
...
+	engine.Use(gin.LoggerWithConfig(gin.LoggerConfig{
+		SkipPaths: []string{"/health"},
+	}))
+	engine.Use(gin.Recovery())
+
+	configureRequestLimits(engine, cfg)
+	configureMode(cfg)
+	configureBaseRoutes(engine)
+
+	if cfg.RunMode == "localdev" {
```
vikaxsh, line 40: we are checking runmode redundantly can we avoid this and check run mode at one place for all ops

server/routes/router.go
```diff
@@ -1,137 +1,133 @@
...
-	web.Router("/internal/project/:projectid/jobs/:id/clear-destination/recover", etlHandler, "post:RecoverClearDestination")
-	web.Router("/internal/project/:projectid/jobs/:id/statefile", etlHandler, "put:UpdateStateFile")
+	engine.POST("/internal/worker/callback/sync-telemetry", etlHandler.UpdateSyncTelemetry)
+	engine.POST("/internal/project/:projectid/jobs/:id/clear-destination/recover", etlHandler.RecoverClearDestination)
+	engine.PUT("/internal/project/:projectid/jobs/:id/statefile", etlHandler.UpdateStateFile)
 
 	if h.Optimization != nil {
 		optHandler := h.Optimization
+		opt := engine.Group("/api/opt/v1")
+		opt.Use(h.AuthMiddleware())
```
vikaxsh: h.AuthMiddleware() is called twice

server/routes/router.go
```diff
@@ -1,137 +1,133 @@
...
 
 		// tables: view
-		web.Router("/api/opt/v1/:catalog/:database/tables", optHandler, "get:GetTablesWithDetails")
+		opt.GET("/:catalog/:database/tables", optHandler.GetTablesWithDetails)
+	}
+}
 
-		// piggy backing
-		web.Router("/api/opt/v1/*", optHandler, "get:PiggyBacking;post:PiggyBacking;put:PiggyBacking;delete:PiggyBacking")
+type ModuleNoRouteHandler struct {
```
vikaxsh, line 101: can we move the struct and function in httpserver folder

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,174 @@
...
+	engine.StaticFile("/favicon.ico", filepath.Join(frontendDistPath, "favicon.ico"))
+}
+
+func configureNoRoute(engine *gin.Engine, cfg *appconfig.Config, h *handlers.Handler) {
+	moduleHandlers := make([]routes.ModuleNoRouteHandler, 0, 1)
+	if h != nil && h.Optimization != nil {
+		// Register optimization as a module fallback for unmatched /api/opt/v1/*.
+		// This avoids route tree conflicts from wildcard catch-all registration.
+		moduleHandlers = append(moduleHandlers, routes.ModuleNoRouteHandler{
+			PathPrefix: "/api/opt/v1/",
```
vikaxsh: we should place all the routes together in router.go

server/internal/httpserver/server.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/gin-gonic/gin"
+
+	"github.com/datazip-inc/olake-ui/server/internal/appconfig"
+	"github.com/datazip-inc/olake-ui/server/internal/handlers"
+	"github.com/datazip-inc/olake-ui/server/routes"
+)
+
+const frontendDistPath = "/opt/frontend/dist"
+
+type Server struct {
```
vikaxsh, line 21: it would be better if we create all the functions with server as receiver

server/routes/router.go
vishalm0509: Yes, made changes

server/internal/database/source.go
vishalm0509: Screenshot 2026-04-14 at 15 24 14

None of these can be zero value, because in the `BindAndValidate` we are mentioning the fields as required

server/internal/appconfig/appconfig.go
vishalm0509: If env is set then we prioritize **env** over **file**

server/internal/handlers/session.go
vishalm0509: Create makes sense here. Made changes

server/internal/httpserver/server.go
vishalm0509: It will not affect the code. But it's mentioned as good practice. Moved it to the top

server/internal/httpserver/server.go
vishalm0509: Just noticed that `NewHandler` does not even return error. Removed it

server/internal/httpserver/server.go
vishalm0509: We are just reading the value from the already loaded config. All these are separate operations and cannot be grouped together.

server/routes/router.go
vishalm0509: It was called twice because we had 2 separate route groups for **ETL** and **OPT**. Now we are using a root API group where I have applied the middleware. Solved

server/routes/router.go
vishalm0509: Yes we can move it. DOne

vishalm0509: The noModuleConfiguration is better if it's in `router.go` as it's related to that. Move the actual implementation to httpserver package.

server/internal/httpserver/server.go
vishalm0509: yes, moved

server/internal/httpserver/server.go
vishalm0509: Done

server/internal/database/source.go
vikaxsh: still we should use correct function

vikaxsh (approved):

server/conf/app.yaml
```diff
@@ -0,0 +1,30 @@
...
+MAX_UPLOAD_SIZE: 67108864
+
+# Either provide a full DSN here...
+POSTGRES_DB: ""
+# ...or configure via individual parts below.
+OLAKE_POSTGRES_USER: temporal
+OLAKE_POSTGRES_PASSWORD: temporal
+OLAKE_POSTGRES_HOST: postgresql
+OLAKE_POSTGRES_PORT: "5432"
+OLAKE_POSTGRES_DBNAME: postgres
```
hash-data, line 15: have we verified these are being overridden via env vars

server/internal/database/destination.go
```diff
@@ -29,13 +31,16 @@ func (db *Database) CreateDestination(destination *models.Destination) error {
 		return fmt.Errorf("failed to encrypt destination config id[%d]: %s", destination.ID, err)
 	}
 	destination.Config = eConfig
-	_, err = db.ormer.Insert(destination)
-	return err
+	return db.conn.Create(destination).Error
```
hash-data, line 34: how many db connection are being opened

server/internal/database/job.go
```diff
@@ -2,101 +2,79 @@ package database
...
-
+	err := db.conn.
+		Model(&models.Job{}).
+		Select(jobListColumns).
+		Where("project_id = ?", projectID).
+		Preload("Source").
+		Preload("Destination").
+		Preload("CreatedBy").
+		Preload("UpdatedBy").
+		Order("updated_at DESC").
```
hash-data, line 76: let us not use updated at as state also updates the job, which gives inconsistency in page

server/internal/database/job.go
```diff
@@ -155,36 +135,22 @@ func (db *Database) GetJobsByDestinationID(destIDs []int) ([]*models.Job, error)
...
-// BeginTx starts a new transaction
-func (db *Database) BeginTx() (orm.TxOrmer, error) {
-	tx, err := db.ormer.Begin()
-	if err != nil {
-		return nil, fmt.Errorf("failed to begin transaction: %s", err)
-	}
-	return tx, nil
+// UpdateJob updates a job with the given params.
+func (db *Database) UpdateJob(jobID int, params map[string]any) error {
+	return db.conn.Model(&models.Job{}).
```
hash-data, line 151: is it atomic update?

hash-data: also add todo to use context in queries

server/internal/database/database.go
```diff
@@ -1,99 +1,89 @@
...
-	password, _ := web.AppConfig.String(constants.ConfOLakePostgresPassword)
-	host, _ := web.AppConfig.String(constants.ConfOLakePostgresHost)
-	port, _ := web.AppConfig.String(constants.ConfOLakePostgresPort)
-	dbName, _ := web.AppConfig.String(constants.ConfOLakePostgresDBname)
-	sslMode, _ := web.AppConfig.String(constants.ConfOLakePostgresSslmode)
+	user := cfg.OlakePostgresUser
+	password := cfg.OlakePostgresPassword
+	host := cfg.OlakePostgresHost
+	port := cfg.OlakePostgresPort
+	dbName := cfg.OlakePostgresDBName
```
hash-data, line 85: why to create variable use directly

server/internal/database/job.go
```diff
@@ -2,101 +2,79 @@ package database
...
-	// Field names must match struct field names (not database column names)
-	_, err := db.ormer.QueryTable(constants.TableNameMap[constants.JobTable]).
-		Filter("project_id", projectID).
-		RelatedSel().
-		OrderBy(constants.OrderByUpdatedAtDesc).
-		All(&jobs, JobListFields...)
-
+	err := db.conn.
+		Model(&models.Job{}).
+		Select(jobListColumns).
```
hash-data, line 70: now we are using snake case ? is that fine?

server/internal/database/user.go
```diff
@@ -11,43 +14,67 @@ import (
...
 	return bcrypt.CompareHashAndPassword([]byte(hashedPassword), []byte(plainPassword))
 }
 
 func (db *Database) CreateUser(user *models.User) error {
-	exists := db.ormer.QueryTable(constants.TableNameMap[constants.UserTable]).Filter("username", user.Username).Exist()
-	if exists {
-		return fmt.Errorf("username already exists")
+	var count int64
+	if err := db.conn.Model(&models.User{}).Where("username = ?", user.Username).Count(&count).Error; err != nil {
+		return err
```
hash-data, line 34: we should wrap error properly so that it is easy to debug

server/internal/database/user.go
```diff
@@ -11,43 +14,67 @@ import (
...
 	}
-
-	_, err := db.ormer.Insert(user)
-	return err
+	if count > 0 {
+		return fmt.Errorf("%w", constants.ErrUserAlreadyExists)
```
hash-data, line 37: error formatting should be correct

server/internal/handlers/session.go
```diff
@@ -0,0 +1,83 @@
...
+
+func newSessionStore(cfg *appconfig.Config, db *database.Database) *sessionStore {
+	return &sessionStore{
+		db:      db,
+		enabled: cfg.SessionOn,
+		secure:  cfg.RunMode != "localdev",
+	}
+}
+
+func (s *sessionStore) SetUserSession(c *gin.Context, userID int) error {
```
hash-data, line 39: why we are managing session ourselves, is there nothing provided inbuilt?

server/internal/handlers/session.go
```diff
@@ -0,0 +1,83 @@
...
+
+	rawPayload, err := s.db.GetActiveSessionData(sessionID)
+	if err != nil {
+		return 0, false
+	}
+
+	var payload sessionPayload
+	if err := json.Unmarshal(rawPayload, &payload); err != nil || payload.UserID == 0 {
+		return 0, false
+	}
```
hash-data, line 80: why error is ignored? we must print

server/conf/app.yaml
vishalm0509: Yes, env vars have higher priority if set

server/internal/database/job.go
vishalm0509: > is it atomic update?
It is a single operation right, so it is atomic only. 

Shall I add context in this PR ?

server/internal/database/database.go
vishalm0509: Yes, done

server/internal/database/job.go
vishalm0509: camelCase is used only in compaction APIs right

server/internal/database/user.go
vishalm0509:
```go
	return nil, fmt.Errorf("failed to get user: %s", err)
```

we are wrapping in the service layer

server/internal/database/user.go
vishalm0509:
```go
errors.Is(err, constants.ErrUserAlreadyExists)
```

we are reading the error in the service layer using `errors.Is()` so we need to format with `%w` only

server/internal/handlers/session.go
vishalm0509: There's no inbuilt. We can use `https://github.com/gin-contrib/sessions` this as well.

server/internal/handlers/session.go
vishalm0509: Done

server/internal/database/destination.go
vishalm0509: The `MaxOpenConnection` is set to `0` so there's no limit on open connections. 
Right, now there's one **idle** and one **inUse** (total = 2) connections open.

server/internal/database/job.go
vishalm0509: Added a TODO

server/internal/database/job.go
vishalm0509: This will ensure the last run jobs will come to the first right. Isn't it better ? Will discuss with Product team ?

hash-data: the thing is once we go to logs and come back we always have to find the job again

server/internal/database/destination.go
hash-data: what was previously?

server/internal/database/job.go
vishalm0509: We are handling this later right

server/internal/database/job.go
vishalm0509: - In old Beego code, fields were Go struct names `("ID", "CreatedBy")`.
- In current GORM code, select list uses DB column names `("id", "created_by_id")`.

server/internal/handlers/session.go
vishalm0509: Should we use the gin-contrib ? if so we can add a todo?

server/internal/database/destination.go
vishalm0509: previously we did not set, similarly we are not setting it now.

- Now default of `MaxOpenConnection=0` -> No limit
- Earlier in Beego as well that's the case: https://pkg.go.dev/database/sql#DB.SetMaxOpenConns

PR #336, opened 2026-03-15
chore: upgrade go version from 1.24.13 to 1.25.8

vishalm0509:
# Description

Upgrade go version from 1.24.13 to 1.25.8

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


.github/workflows/golang-ci.yml
```diff
@@ -21,11 +21,11 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "server/go.mod"
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
+          install-mode: goinstall
```
deepanshupal09-datazip, line 28: The default installation mode for `golangci/golangci-lint-action` is binary, which downloads a prebuilt binary and is generally faster and more stable in CI. Using `install-mode: goinstall` forces compilation from source, which can slow down the workflow and introduce potential build issues without a clear benefit here.

Unless there was a specific reason for switching to `goinstall` (e.g., binary incompatibility with the Go version), it makes sense to drop this option and rely on the default behavior.

> AI review, please check it's relevance.

vishalm0509: The default mode uses cached binary which does not guarantee using the go-version we specify. So there was a conflict of version mismatch and CI fails

deepanshupal09-datazip (dismissed):
LGTM

.github/workflows/golang-ci.yml
```diff
@@ -18,25 +18,24 @@ jobs:
...
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "server/go.mod"
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v6
         with:
+          install-mode: goinstall
           version: latest
```
deepanshupal09-datazip, line 29: while we are at it, we should also fix the linter version(instead of using latest) everywhere so that CI doesn't fail on the code that hasn't even changed.

vishalm0509: We can change it when it starts failing. Confirmed with the team

deepanshupal09-datazip (approved):

PR #376, opened 2026-05-12
chore: add configurable http timeouts

vishalm0509:
# Description

Hardcoded HTTP server timeouts are now configurable. HTTP server timeouts are now set in app.yaml (and can be overridden with env vars). docker-compose-v1.yml uses the same longer timeouts.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


server/internal/httpserver/server.go
```diff
@@ -49,10 +49,10 @@ func New(cfg *appconfig.Config, h *handlers.Handler) *Server {
...
 		Addr:              fmt.Sprintf(":%s", cfg.HTTPPort),
 		Handler:           s.engine,
-		ReadHeaderTimeout: 10 * time.Second,
-		ReadTimeout:       30 * time.Second,
-		WriteTimeout:      60 * time.Second,
-		IdleTimeout:       120 * time.Second,
+		ReadHeaderTimeout: cfg.HTTPReadHeaderTimeout,
+		ReadTimeout:       cfg.HTTPReadTimeout,
+		WriteTimeout:      cfg.HTTPWriteTimeout,
+		IdleTimeout:       cfg.HTTPIdleTimeout,
```
deepanshupal09-datazip, line 55: please add zero-value guards for the timeouts in `loadConfig()` since `viper.GetDuration()` silently returns 0 on a bad env var, and 0 = no timeout in net/http.

vishalm0509: - We are providing application per customer right, so even if set to 0, resource contention will not be a problem. 

- We also allow setting zero as a value as that's the default value if the timeout property is not set. 

Let's not validate zero value.

deepanshupal09-datazip (approved):

PR #378, opened 2026-05-15
feat: bulk edit for table configurations and go security fix

arsalan-datazip:
# Description
Earlier user can only configure one table at a time and enable its optimization.
This PR adds a bulk edit API wihich allows user to configure multiple tables and enable optimisation of  all selected tables.

## Type of change


server/routes/router.go
```diff
@@ -91,6 +91,7 @@ func RegisterRoutes(engine *gin.Engine, h *handlers.Handler) {
 		opt.DELETE("/catalog/:catalog", optHandler.DeleteCatalog)
 
 		// terminal: cron, enable/disable optimization
+		opt.PUT("/:catalog/:database/config/bulk", optHandler.SetBulkProperties)
```
badalprasadsingh: For a single table its: `/:catalog/:database/:table/config`
So, for multiple tables keep it: `/:catalog/:database/tables/config`

server/internal/constants/constants.go
```diff
@@ -101,7 +101,7 @@ var (
 	OptPathTerminalExecute          = "/api/ams/v1/terminal/catalogs/%s/execute"
 	OptPathTerminalLogs             = "/api/ams/v1/terminal/%s/logs"
 	// others
-	OptMaxTimeout          = 30 * time.Second
+	OptMaxTimeout          = 5 * time.Minute
```
badalprasadsingh: This constant is also used for Optimization client, in `http.Client`. It shouldn't be as long as 5 minutes there.

Please create a new constant, specially for Bulk Feature (given, such a constant is really required)

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
-func (s *Service) SetTableProperties(ctx context.Context, req dto.SetTablePropertiesRequest) (*dto.TableProperties, error) {
-	var propsSQL []string
-	for key, value := range req.Properties {
-		propsSQL = append(propsSQL, fmt.Sprintf("'%s' = '%s'", key, value))
+// createAlterQuery returns one ALTER TABLE ... SET TBLPROPERTIES (...) statement for database.table, ending with ';'.
+func createAlterQuery(database, table string, properties map[string]string) string {
+	keys := make([]string, 0, len(properties))
+	for k := range properties {
+		keys = append(keys, k)
+	}
```
badalprasadsingh: Why make an unnecessary list of keys here?

server/internal/services/optimization/terminal.go
```diff
@@ -13,9 +13,8 @@ import (
 	"github.com/datazip-inc/olake-ui/server/internal/utils"
 )
 
-func (s *Service) SetProperties(ctx context.Context, catalog, database, table string, config dto.SQLInput) (*dto.TableProperties, error) {
+func SetPropertiesMap(config dto.SQLInput) map[string]string {
```
badalprasadsingh: We are not exposing this function. Make the first letter small.

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+	for _, k := range keys {
+		props = append(props, fmt.Sprintf("'%s'='%s'", k, properties[k]))
+	}
+	propsJoined := strings.Join(props, ", ")
+
+	return fmt.Sprintf(constants.OptSQLCommand, database, table, propsJoined) + ";"
+}
+
+// BulkSetProperties runs one AMS terminal session that executes ALTER TABLE ... SET TBLPROPERTIES
+// for every table in req.Tables (same property set on each).
```
badalprasadsingh: Avoid using these keywords: "AMS"

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+func (s *Service) BulkSetProperties(ctx context.Context, catalog, database string, req dto.BulkSQLInput) (*dto.BulkTableProperties, error) {
+	tables := req.Tables
+	properties := SetPropertiesMap(req.SQLInput)
+	// Bulk apply always enables self-optimizing on every selected table.
+	properties[constants.OptEnableOptimization] = "true"
```
badalprasadsingh: Is it written in PRD?

Isn't it in the use case where: just cron configurations are set for multiple tables in a single API call?

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+
+	sql := make([]string, 0, len(tables))
+	for _, t := range tables {
+		sql = append(sql, createAlterQuery(database, t, properties))
+	}
+	// One AMS terminal execute body: multiple statements, one Spark session (AMS splits on ';' per line).
```
badalprasadsingh: Avoid "AMS"

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+	path := fmt.Sprintf(constants.OptPathTerminalExecute, catalog)
+	requestBody := dto.TerminalExecuteRequest{
+		SQL: bulkSQLScript,
 	}
 
-	sql := fmt.Sprintf(constants.OptSQLCommand, req.Database, req.Table, strings.Join(propsSQL, ", "))
+	var sessionResult dto.TerminalSessionResponse
+	if err := s.DoInto(ctx, http.MethodPost, path, url.Values{}, requestBody, &sessionResult); err != nil {
+		return nil, fmt.Errorf("failed to execute bulk ALTER TABLE for catalog %s: %w", catalog, err)
+	}
```
badalprasadsingh: also return the database name in error

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+
+	logInfo, err := s.pollForCompletion(ctx, catalog, sessionResult.SessionID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to poll for completion: %w", err)
+	}
```
badalprasadsingh: This error statement is same as for `SetTableProperties`. We should have unique error statements for faster debugging.

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,80 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+		message = fmt.Sprintf(
+			"bulk optimization sql command completed successfully for %d table(s), session ID: %s",
+			len(tables),
+			sessionResult.SessionID,
+		)
+	} else {
+		message = fmt.Sprintf(
+			"bulk optimization sql command failed for catalog %s, session ID: %s",
+			catalog,
+			sessionResult.SessionID,
```
badalprasadsingh: please return the database name as well

badalprasadsingh (changes requested):
Reviewed.

Just to confirm: this feature is for multiple tables across a single database, and not across multiple databases as well, right?

server/internal/services/optimization/terminal.go
arsalan-datazip: for all those tables set in bulk edit we have to enable the optimization for all

server/internal/services/optimization/terminal.go
arsalan-datazip: done

server/internal/services/optimization/terminal.go
arsalan-datazip: done

server/routes/router.go
arsalan-datazip: updated

server/internal/constants/constants.go
arsalan-datazip: done, added new different constant for terminal session timeout

server/internal/services/optimization/terminal.go
arsalan-datazip: removed

server/internal/services/optimization/terminal.go
arsalan-datazip: done

server/internal/services/optimization/terminal.go
arsalan-datazip: done

server/internal/constants/constants.go
```diff
@@ -102,6 +102,7 @@ var (
 	OptPathTerminalLogs             = "/api/ams/v1/terminal/%s/logs"
 	// others
 	OptMaxTimeout          = 30 * time.Second
+	OptSessionTimeout      = 1 * time.Minute
```
badalprasadsingh: Since `OptSessionTimeout` is set at max 1 min, thus the polling function will wait for a maximum of 1 min.

For Bulk, the user might select 1000 tables. Are we sure it will finish in one minute?. Please discuss it with Ankit once.

server/internal/models/dto/requests.go
```diff
@@ -181,3 +181,9 @@ type SQLInput struct {
 	TargetFileSize         *int64  `json:"target_file_size"`
 	EnabledForOptimization *string `json:"enabled_for_optimization"`
 }
+
+// BulkSQLInput configures optimization properties for multiple tables in one AMS terminal session.
```
badalprasadsingh: Please avoid "AMS" wherever used.

server/go.mod
```diff
@@ -174,10 +174,10 @@ require (
...
 	github.com/subosito/gotenv v1.6.0 // indirect
 	go.temporal.io/api v1.62.0
-	golang.org/x/net v0.53.0 // indirect
+	golang.org/x/net v0.54.0 // indirect
 	golang.org/x/sync v0.20.0 // indirect
-	golang.org/x/sys v0.43.0 // indirect
-	golang.org/x/text v0.36.0 // indirect
+	golang.org/x/sys v0.45.0 // indirect
+	golang.org/x/text v0.37.0 // indirect
 	golang.org/x/time v0.14.0 // indirect
```
badalprasadsingh: Why do we need to change these versions?

badalprasadsingh (changes requested):
Second Review. Will test after this.

Single Table configuration Service Function is changed as well, please do check backward compatibility for that API.

server/go.mod
arsalan-datazip: To fix go vulnerability error.

arsalan-datazip:
> Second Review. Will test after this.
> 
> Single Table configuration Service Function is changed as well, please do check backward compatibility for that API.



Will do backward compatibilty check
I have changed single Table conf because both SetProperties and BulkSetProperties were using the same logic to create map for properties so just to abstract it out I kept that in a single function .

server/internal/services/optimization/terminal.go
```diff
@@ -94,7 +157,7 @@ func (s *Service) SetTableProperties(ctx context.Context, req dto.SetTableProper
 // pollForCompletion polls the terminal API for SQL execution completion
 func (s *Service) pollForCompletion(ctx context.Context, _, sessionID string) (*dto.LogInfo, error) {
```
badalprasadsingh: `pollForCompaction` function takes an extra unnecessary argument `_` argument, which is just not required in the function.

Callers of this function pass catalog name. Please take care of it in this PR itself.

badalprasadsingh (changes requested):

badalprasadsingh:
This PR also fixes "Go Security Fix". Please add that in the title as well.

server/internal/models/dto/response.go
```diff
@@ -252,6 +252,15 @@ type TableProperties struct {
...
 
+// BulkTableProperties is the response from bulk-setting optimization properties.
+type BulkTableProperties struct {
+	SessionID string   `json:"sessionId"`
+	Success   bool     `json:"success"`
+	Message   string   `json:"message"`
+	Logs      []string `json:"logs,omitempty"`
+	Tables    []string `json:"tables"`
+}
+
```
badalprasadsingh: We already have a dto of `TableProperties`. BulkTableProperties response is same as TableProperties, only addition is the tables sent back. I think we can remove this dto, as sending that same tables back as response seems unnecessary.

Basically, we can reuse the dto.

server/internal/services/optimization/terminal.go
```diff
@@ -31,7 +30,11 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
 		properties[constants.OptTargetFileSize] = utils.ConvertMBToBytes(*config.TargetFileSize)
 	}
+	return properties
+}
 
+func (s *Service) SetProperties(ctx context.Context, catalog, database, table string, config dto.SQLInput) (*dto.TableProperties, error) {
+	properties := setPropertiesMap(config)
 	// sql query
 	sqlResult, err := s.SetTableProperties(ctx, dto.SetTablePropertiesRequest{
 		Catalog:    catalog,
```
badalprasadsingh: This `SetProperties` service function is ultimately calling another service function `SetTableProperties`, which seems uncessary. Previously, it was required as we did not have `setPropertiesMap` function.

Please club them into a single service function, `SetProperties`.

server/internal/services/optimization/terminal.go
```diff
@@ -47,14 +50,73 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+	for k, value := range properties {
+		props = append(props, fmt.Sprintf("'%s'='%s'", k, value))
 	}
+	propsJoined := strings.Join(props, ", ")
+
+	return fmt.Sprintf(constants.OptSQLCommand, database, table, propsJoined) + ";"
+}
 
-	sql := fmt.Sprintf(constants.OptSQLCommand, req.Database, req.Table, strings.Join(propsSQL, ", "))
+// BulkSetProperties runs one terminal session that executes ALTER TABLE ... SET TBLPROPERTIES
```
badalprasadsingh: Keep it as per logic and simple like: "sets multiple table properties using the SQL query"

badalprasadsingh (changes requested):
Few more comments.

badalprasadsingh:
image


Please return an error message in case the user does not selects any table.

This behaviour is required according to the PRD.

server/internal/services/optimization/terminal.go
hash-data: let us discuss, i think there might be option to enable for all

server/internal/services/optimization/terminal.go
```diff
@@ -31,56 +30,99 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database, table st
...
+		message = fmt.Sprintf(
+			"bulk optimization sql command completed successfully for %d table(s), session ID: %s",
+			len(tables),
+			sessionResult.SessionID,
+		)
 	} else {
-		message = fmt.Sprintf("optimization sql command failed with session ID: %s", sessionResult.SessionID)
+		message = fmt.Sprintf(
+			"bulk optimization sql command failed for catalog %s, database %s, session ID: %s",
+			catalog,
```
hash-data: code get duplicated can we check this?

hash-data (dismissed):

hash-data (approved):

badalprasadsingh (approved):
LGTM.

PR #379, opened 2026-05-18
feat: add support for external temporal

schitizsharma:
# Description
Adds support for external Temporal in the olake-ui server, mirroring the implementation in `olake-helm`. Previously the server only supported bundled Temporal with a hardcoded task queue. This change enables three deployment modes driven entirely by environment variables:
  - **Default**: `TEMPORAL_EXTERNAL=false` — no change in behavior, connects to the bundled Temporal service at `TEMPORAL_ADDRESS` using `OLAKE_DOCKER_TASK_QUEUE`
  - **External self-managed**: `TEMPORAL_EXTERNAL=true`, `TEMPORAL_API_KEY` unset — connects to a user-managed Temporal cluster with configurable namespace and task queue
  - **Temporal Cloud**: `TEMPORAL_EXTERNAL=true`, `TEMPORAL_API_KEY` set — connects to Temporal Cloud with TLS and API key credentials

Fixes # (issue)

## Type of change

# How Has This Been Tested?
  - [x] **Default**: Start with default config (`TEMPORAL_EXTERNAL` unset), verify server connects to bundled Temporal and workflows execute normally on `OLAKE_DOCKER_TASK_QUEUE`
  - [x] **External self-managed**: Set `TEMPORAL_EXTERNAL=true`, `TEMPORAL_ADDRESS=<host:port>`, `TEMPORAL_NAMESPACE=<ns>`, `TEMPORAL_TASK_QUEUE=<queue>` — verify server connects to external
  cluster and dispatches workflows to the correct queue
  - [x] **Temporal Cloud**: Set `TEMPORAL_EXTERNAL=true`, `TEMPORAL_API_KEY=<key>`, `TEMPORAL_ENABLE_TLS=true`, `TEMPORAL_NAMESPACE=<ns>`, `TEMPORAL_TASK_QUEUE=<queue>` — verify TLS + API key
   credentials are injected and connection succeeds
  - [x] Backward compatibility with current latest release of olake-ui

# Screenshots or Recordings
  

## Related PR's (If Any):
  - olake-helm: feat/worker-external-temporal (worker-side implementation this aligns with)


server/internal/services/temporal/client.go
```diff
@@ -21,23 +22,42 @@ type Temporal struct {
...
-			HostPort: temporalAddress,
-		})
+		clientOptions := client.Options{
+			HostPort:  cfg.TemporalAddress,
+			Namespace: cfg.TemporalNamespace,
+		}
+
+		if cfg.TemporalEnableTLS {
+			clientOptions.ConnectionOptions = client.ConnectionOptions{
+				TLS: &tls.Config{}, // #nosec G402 -- Temporal SDK handles TLS negotiation internally
```
vishalm0509: Can we force a min version here, instead of supressing it. 

Anyways all lower versions are deprecated and has vulnerabilities: https://learn.microsoft.com/en-us/dotnet/framework/network-programming/tls?tabs=47-plus%2Chttpclient-sslstream#what-is-transport-layer-security-tls

server/internal/services/temporal/client.go
```diff
@@ -21,23 +22,42 @@ type Temporal struct {
...
 	err := utils.RetryWithBackoff(func() error {
-		client, dialErr := client.Dial(client.Options{
-			HostPort: temporalAddress,
-		})
+		clientOptions := client.Options{
+			HostPort:  cfg.TemporalAddress,
+			Namespace: cfg.TemporalNamespace,
+		}
+
+		if cfg.TemporalEnableTLS {
```
vishalm0509, line 43: We need to set it for cloud as well ?

```suggestion
		if cfg.TemporalEnableTLS || cfg.TemporalAPIKey != "" {
```

server/conf/app.yaml
```diff
@@ -25,6 +25,11 @@ LOGS_DIR: ./logger/logs
 SESSION_ON: true
 
 TEMPORAL_ADDRESS: temporal:7233
+TEMPORAL_NAMESPACE: default
+TEMPORAL_ENABLE_TLS: false
+TEMPORAL_API_KEY: ""
+TEMPORAL_EXTERNAL: false
+TEMPORAL_TASK_QUEUE: ""
```
vishalm0509, line 32: These should we added to `docker-compose-v1` in the shared envs so that it is read by both bff and worker

server/internal/services/temporal/client.go
schitizsharma: TLS is handled by the temporal SDK itself, point is that if Temporal decides to move forward to a newer TLS version the client's environment may break.
We will just simply let Temporal SDK handle it internally.

server/internal/services/temporal/client.go
schitizsharma: Ah! so you can have an external temporal with HTTPs encryption which is not cloud. 
For example, like one of our initial client's asked, they could have their own self-managed temporal cluster internally which they want to hook into OLake stack.

Temporal API Key is only for cloud, it is not a feature of Temporal Open Source.

server/conf/app.yaml
schitizsharma: Good point! Let's add them when this PR is ready for merge.
Keep this comment open and only resolve it once the docker-compose-v1 side implementation done.

server/internal/services/temporal/client.go
vishalm0509: https://docs.temporal.io/self-hosted-guide/security#encryption-in-transit-with-mtls

But it is mentioned that TLS is supported for both CLoud and Self-hosted versions

server/internal/services/temporal/client.go
vishalm0509: But we can mention the minVersion, and if temporal bumps the minversion up it will not affect the environment (that happens in case of MaxVersion). 

So if we set, we can remove the suppression as well.

schitizsharma: Ok, I will update it on both `olake-ui` and `olake-helm` PRs

vishalm0509 (approved):
LGTM

Issue #380, opened 2026-05-21
Docker Hub coupling makes the UI unusable for Fusion-only / air-gapped deployments


antoniocali:
## Summary

The UI server hard-couples several code paths to `hub.docker.com` (the metadata API, not the registry), which is fragile in any deployment where outbound to Docker Hub is rate-limited, slow, or restricted — and is **dead weight** in Fusion-only deployments that never run CDC sync jobs.

I'm running OLake on Kubernetes purely as the Iceberg maintenance layer (Fusion + Spark optimizer) against an existing Polaris catalog. No `olakeWorker` CDC, no source / destination connectors. Observed three classes of issue:

### 1. Continuous background polling generates log noise

Every ~30s the UI polls Docker Hub for tags on every CDC source connector image, even when CDC is unused:

```
2026-05-21T15:41:22Z WRN failed to fetch image tags online for olakego/source-mysql:
  docker hub api request failed with status code: 504.
  Cached fallback unavailable on Kubernetes (no Docker daemon)
2026-05-21T15:41:52Z WRN failed to fetch image tags online for olakego/source-postgres: …
2026-05-21T15:42:22Z WRN failed to fetch image tags online for olakego/source-oracle: …
2026-05-21T15:42:52Z WRN failed to fetch image tags online for olakego/source-mongodb: …
… (also kafka, s3, db2, mssql)
```

Eight connectors × ~30s polling = ~16 calls/min hitting Docker Hub. With the (very strict) anonymous metadata-API rate limits, these mostly 504 and just generate continuous warnings.

The `Cached fallback unavailable on Kubernetes (no Docker daemon)` text is also confusing — it implies a feature is missing on K8s rather than that the deployment doesn't need it.

### 2. User-facing endpoints become slow / fail when Docker Hub throttles

When the rate limit is in effect, user-facing endpoints that *also* call Docker Hub become very slow or 500:

```
[GIN] 2026/05/21 - 15:36:48 | 500 |   4m0s    | GET  /api/v1/project/123/destinations/versions?type=iceberg
[GIN] 2026/05/21 - 15:38:55 | 500 |   30.1s   | GET  /api/v1/project/123/sources/versions?type=mongodb
[GIN] 2026/05/21 - 15:40:49 | 500 |   4m0s    | GET  /api/v1/project/123/destinations/versions?type=iceberg
[GIN] 2026/05/21 - 20:58:23 | 200 |  15.58s   | POST /api/v1/project/123/destinations/spec
```

[...]

PR #382, opened 2026-05-22
feat(server): add CONNECTOR_DISCOVERY_ENABLED flag to disable registry polling

antoniocali:
## Summary

Closes part of #380.

Adds a new server config flag `CONNECTOR_DISCOVERY_ENABLED` (default `true`) that, when disabled, short-circuits `utils.GetDriverImageTags` and returns a configurable `DEFAULT_CONNECTOR_VERSION` instead of querying Docker Hub / ECR / GCR Artifact Registry. Default-on preserves the current behavior for every existing deployment.

This addresses the Fusion-only / air-gapped use case described in #380, where the UI hits Docker Hub on every `/versions`, `/spec`, and `/test-connection` call even though CDC connectors are unused. With discovery disabled, those endpoints return the configured version without an outbound HTTP request, eliminating both the user-facing slowness during rate-limited periods and the log noise.

This is suggestion (1) from the issue. Suggestion (5) (rewording the K8s warning) is bundled here since it's a one-liner and naturally points operators at the new flag. Suggestions (2) backoff, (3) DB spec cache, and (4) custom discovery registry are intentionally out of scope.

## Changes

- `server/internal/appconfig/appconfig.go` — two new fields on `Config`: `ConnectorDiscoveryEnabled` (bool) and `DefaultConnectorVersion` (string).
- `server/conf/app.yaml` — defaults: `CONNECTOR_DISCOVERY_ENABLED: true`, `DEFAULT_CONNECTOR_VERSION: ""`. Behavior unchanged unless an operator explicitly opts in.
- `server/internal/utils/docker_utils.go` — `GetDriverImageTags` short-circuits at the top of the function when discovery is disabled; returns `(nil, "", error)` with a clear message if `DEFAULT_CONNECTOR_VERSION` is unset.
- `server/internal/utils/docker_utils.go` — rewords the Kubernetes "Cached fallback unavailable on Kubernetes (no Docker daemon)" warning at line 103 to be less alarming and to mention the new flag.

## Helm chart

A separate docs-only PR to `datazip-inc/olake-helm` will surface the flag via `olakeUI.env` examples + a README note. No template change is required there because `olakeUI.env` is already a passthrough map.

## Operator usage

```yaml
# values.yaml
olakeUI:
  env:
    CONNECTOR_DISCOVERY_ENABLED: "false"
    DEFAULT_CONNECTOR_VERSION: "v0.3.18"
```

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [ ] Manual: verify default (`CONNECTOR_DISCOVERY_ENABLED=true`) preserves existing `/versions` and `/spec` behavior
- [ ] Manual: verify `CONNECTOR_DISCOVERY_ENABLED=false` with `DEFAULT_CONNECTOR_VERSION=v0.3.18` makes `/versions` return `["v0.3.18"]` without any outbound HTTP
- [ ] Manual: verify `CONNECTOR_DISCOVERY_ENABLED=false` with empty `DEFAULT_CONNECTOR_VERSION` returns a clear error on `/versions` and `/spec`

### DIFF
--- a/server/conf/app.yaml
+++ b/server/conf/app.yaml
@@ -27,6 +27,16 @@ SESSION_ON: true
 TEMPORAL_ADDRESS: temporal:7233
 CONTAINER_REGISTRY_BASE: registry-1.docker.io
 
+# Set to false in Fusion-only or air-gapped deployments to disable upstream
+# container registry queries (Docker Hub / ECR / GCR) for connector version
+# discovery. When false, /versions and /spec return DEFAULT_CONNECTOR_VERSION
+# instead of polling the registry.
+CONNECTOR_DISCOVERY_ENABLED: true
+
+# Fallback connector version used when CONNECTOR_DISCOVERY_ENABLED is false.
+# Ignored when discovery is enabled. Example: "v0.3.18".
+DEFAULT_CONNECTOR_VERSION: ""
+
 # Prefer setting this via environment.
 OLAKE_SECRET_KEY: ""
 
--- a/server/internal/appconfig/appconfig.go
+++ b/server/internal/appconfig/appconfig.go
@@ -35,6 +35,17 @@ type Config struct {
 	OptimizationBaseURL   string
 	OptimizationUsername  string
 	OptimizationPassword  string
+
+	// ConnectorDiscoveryEnabled controls whether the server queries the
+	// upstream container registry (Docker Hub / ECR / GCR Artifact Registry)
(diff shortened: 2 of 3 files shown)


nayanj98:
@antoniocali we are looking into this issue and your PR. Will have an update on it soon.

antoniocali:
any news on this @nayanj98 ?

PR #386, opened 2026-06-03
feat(opt): catalog test connection and spec api

arsalan-datazip:
# Description

This PR adds the test connection API for catalogs that will check the credentials/configuration before creating the catalog

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


resources/spec.json
badalprasadsingh, line 1: Please keep this file in this path. `server/internal/services/optimization/resources/spec.json`

Dockerfile
```diff
@@ -58,6 +58,7 @@ RUN mkdir -p conf config /opt/frontend/dist
 # Copy built artifacts from builder stages
 COPY --from=go-builder /app/olake-server ./olake-server
 COPY server/conf/app.yaml ./config/app.yaml
+# COPY resources/spec.json ./resources/spec.json
```
badalprasadsingh: Why this is commented out?

server/routes/router.go
```diff
@@ -85,6 +85,8 @@ func RegisterRoutes(engine *gin.Engine, h *handlers.Handler) {
 		opt := api.Group("/opt/v1")
 
 		// catalogs: crud
+		opt.POST("/catalog/spec", optHandler.GetCatalogSpec)
+		opt.POST("/catalog/test", optHandler.TestCatalogConnection)
 		opt.POST("/catalog", optHandler.CreateCatalog)
 		opt.GET("/catalog/:catalog", optHandler.GetCatalog)
 		opt.PUT("/catalog/:catalog", optHandler.UpdateCatalog)
```
badalprasadsingh, line 91: nit:

```go
		// catalogs
		opt.POST("/catalog/spec", optHandler.GetCatalogSpec)
		opt.POST("/catalog/test", optHandler.TestCatalogConnection)
		// catalogs: crud
		opt.POST("/catalog", optHandler.CreateCatalog)
		opt.GET("/catalog/:catalog", optHandler.GetCatalog)
		opt.PUT("/catalog/:catalog", optHandler.UpdateCatalog)
		opt.DELETE("/catalog/:catalog", optHandler.DeleteCatalog)
```

server/internal/constants/constants.go
```diff
@@ -118,6 +119,9 @@ var (
 	// hard-coding to S3 now, as the other options are "hadoop" & "OSS" for optimization
 	// GCS & ADLS are supported, given the catalog manages the sdk (eg, Lakekeeper with GCS flavour)
 	DefaultOptimizationStorageType = "S3"
+	CatalogSpecVersion             = "v0.2.7"
```
badalprasadsingh: v0.2.7? or v0.7.2?

badalprasadsingh (changes requested):
Reviewed. Few more comments.

server/internal/constants/constants.go
arsalan-datazip: v0.2.7

server/internal/handlers/optimization/catalog.go
```diff
@@ -112,3 +143,23 @@ func (h *Handler) requiredCatalog(c *gin.Context) (string, bool) {
...
+	if err != nil {
+		utils.ErrorResponse(c, http.StatusInternalServerError, "failed to read catalog spec", err)
+		return
+	}
+
+	var spec map[string]interface{}
+	if err := json.Unmarshal(data, &spec); err != nil {
+		utils.ErrorResponse(c, http.StatusInternalServerError, "invalid catalog spec JSON", err)
+		return
+	}
```
badalprasadsingh, line 119: Avoid having any product level logic in handlers. Handlers should always be transparent.

Create a service function and call that from this handler.

badalprasadsingh:
Change the PR title to `feat(opt): catalog connection test and spec api`

server/internal/constants/constants.go
```diff
@@ -118,6 +119,9 @@ var (
...
+	CatalogSpecType                = "iceberg"
+	IcebergCatalogSpecFile         = "resources/spec.json"
```
badalprasadsingh: When you create the OLake-UI image, check if this path is valid inside the image and the API works. Test this.

server/routes/router.go
```diff
@@ -85,6 +85,8 @@ func RegisterRoutes(engine *gin.Engine, h *handlers.Handler) {
 		opt := api.Group("/opt/v1")
 
 		// catalogs: crud
+		opt.POST("/catalog/spec", optHandler.GetCatalogSpec)
```
badalprasadsingh: Why is `GetCatalogSpec` POST?

badalprasadsingh: Also after you convert it to GET, it will collide with GET catalog/:catalog (if the catalog name is spec itself)

Better make it:
`/catalog/resources/spec`

badalprasadsingh (changes requested):
Please check these as well.

server/internal/constants/constants.go
badalprasadsingh: Why `v0.2.7`? It should return OLake version. Look at the current version returned in the destination spec API.

Add the OLake version whose spec you have added.

resources/spec.json
badalprasadsingh: Keep the file in services, exactly as mentioned above

server/internal/constants/constants.go
arsalan-datazip: Tested, API is working and returning the hardcoded spec

server/internal/services/optimization/resources/spec.json
badalprasadsingh, line 1: This is not same as the response of spec from `v0.7.4`. Please check.

badalprasadsingh (changes requested):
Reviewed again. One comment.

server/internal/services/optimization/client.go
```diff
@@ -268,3 +266,13 @@ func getCredentials() (string, string, string, error) {
 
 	return baseURL, username, password, nil
 }
+
+
+func (s *Service) clientForTimeout(path string) {
```
badalprasadsingh: nit: please rename the function to `setClientTimeout`

server/internal/services/optimization/client.go
```diff
@@ -268,3 +266,12 @@ func getCredentials() (string, string, string, error) {
...
 }
+
+func (s *Service) setClientTimeout(path string) {
+	switch path {
+	case constants.OptPathCatalogTest:
+		s.client.Timeout = constants.OptTestCatalogTimeout
+	default:
+		s.client.Timeout = constants.OptMaxTimeout
+	}
+}
```
badalprasadsingh: There is a major race condition here. Since, we are changing the the parameter of `s.client.Timeout`:

1. API A running with 30s client
2. API B starts with 10 mins client
3. API B makes the client timeout to be 10 mins, so API A also has 10 mins timeout now.

badalprasadsingh: Make it like this in `sendRequest` function:

```go
	var bodyReader io.Reader
	if bodyBytes != nil {
		bodyReader = bytes.NewReader(bodyBytes)
	}

    // this
	ctx, cancel := context.WithTimeout(ctx, requestTimeout(path))
	defer cancel()

	req, err := http.NewRequestWithContext(ctx, method, fullURL, bodyReader)
	if err != nil {
		return nil, 0, nil, fmt.Errorf("failed to create request to optimization: %s", err)
	}
```

badalprasadsingh:
```go
func requestTimeout(path string) (time.Duration) {
	switch path {
	case constants.OptPathCatalogTest:
		return constants.OptTestCatalogTimeout
	default:
		return constants.OptMaxTimeout
	}
}
```

server/internal/constants/constants.go
```diff
@@ -102,6 +103,7 @@ var (
 	OptPathTerminalLogs             = "/api/ams/v1/terminal/%s/logs"
 	// others
 	OptMaxTimeout          = 30 * time.Second
+	OptTestCatalogTimeout  = 10 * time.Minute
```
hash-data: it need to 5 mints only

server/internal/handlers/optimization/catalog.go
```diff
@@ -26,6 +26,33 @@ func (h *Handler) GetCatalog(c *gin.Context) {
 	utils.SuccessResponse(c, "catalog details retrieved successfully", olakeConfig)
 }
 
+func (h *Handler) TestCatalogConnection(c *gin.Context) {
```
hash-data: why we need an api for it?

server/internal/services/optimization/resources/spec.json
```diff
@@ -0,0 +1,299 @@
...
+                    "title": "S3 Endpoint",
+                    "type": "string"
+                  }
+                }
+              }
+            ]
+          }
+        },
+        "properties": {
+          "arrow_writes": {
```
hash-data: we can remove this

server/internal/services/optimization/resources/spec.json
```diff
@@ -0,0 +1,299 @@
...
+            "title": "AWS Region",
+            "type": "string"
+          },
+          "aws_secret_key": {
+            "description": "The AWS secret key for S3 authentication—typically 40+ characters long",
+            "format": "password",
+            "title": "AWS Secret Key",
+            "type": "string"
+          },
+          "catalog_name": {
```
hash-data, line 255: it is required

badalprasadsingh:
**Note:** Included a small fix in this PR: To return `rest_signing_v_4` signature as `false` from Fusion, when UI sets it as false. 

Previously, due to default value in spec `rest_signing_v_4` always showed `true`, even if the key was never passed from backend.

badalprasadsingh (approved):
Server API. Approved.

PR #404, opened 2026-07-13
fix(opt): remove regex in catalog name field from spec and update tooltip

badalprasadsingh:
# Description

1. Removed the pattern regex from `catalog_name` field, allowing even numbers and special characters for catalog name support in Fusion from OLake-UI : Tested through **Glue Catalog**
2. Updated Tooltip

## Type of change

# How Has This Been Tested?

- [x] Tested creating Glue Catalog Names with numbers in it.


server/internal/services/optimization/resources/spec.json
```diff
@@ -253,8 +253,8 @@
             "type": "string"
           },
           "catalog_name": {
-            "description": "User defined catalog name, if left blank defaults to olake_iceberg. Only lowercase letters and underscores are allowed.",
-            "pattern": "^[a-z_]+$",
+            "description": "User defined catalog name. Lowercase letters, numbers and underscores are allowed.",
```
hash-data: can you check other possible accepted patterns other than having special characters?

badalprasadsingh: Done.

Right now, Fusion even supports catalog names with _special characters_, so removed the regex pattern for the field for now.

hash-data (approved):

PR #408, opened 2026-07-24
feat: fusion telemetry

ashisharma-dz:
# Description

Integrated telemetry for fusion in Olake-UI

## Change

- Changes were done in modules where the events were being triggered, including both ui and server
- As well as in analytics/analyticsEnums.ts for definition of events

# How Has This Been Tested?
Live tested by triggering events on the UI and checking the logs.


docker-compose-v1.yml
deepanshupal09-datazip, line 1: please revert all the `docker-compose-v1.yml` file changes, they are not needed.

server/internal/utils/telemetry/constants.go
```diff
@@ -19,4 +19,6 @@ const (
 	EventDestinationCreated  = "destination_created"
 	EventSourcesUpdated      = "sources_updated"
 	EventDestinationsUpdated = "destinations_updated"
+	EventInstalledFusion     = "Installed_Fusion"
+	EventInstallationMode    = "Installation_Mode"
```
deepanshupal09-datazip:
```suggestion
	EventFusionInstalled     = "fusion_installed"
	EventInstallationMode    = "installation_mode"
```
lets keep the naming consistent with the rest of the events.

ui/src/core/layout/components/SidebarNavItem.tsx
```diff
@@ -10,6 +17,11 @@ const SidebarNavItem: React.FC<{
 }> = ({ path, label, icon: Icon, iconSize = 14, className }) => (
 	<NavLink
 		to={path}
+		onClick={() => {
+			if (MAINTENANCE_PATHS.has(path)) {
+				trackEvent(AnalyticsEvent.MaintenanceModuleOpened, { path })
```
deepanshupal09-datazip: `Maintenance_Module_Opened` is fired whenever user navigates inside the maintenance module, lets say he clicks on "Tables" inside side navbar five times, this event will be fired five times... is this expected?

ui/src/core/analytics/analyticsEnums.ts
```diff
@@ -1,7 +1,29 @@
...
-	CreateSourceClicked = "create_source_clicked",
-	CreateDestinationClicked = "create_destination_clicked",
+	TestConnectionSource = "Test_Connection_Source",
+	TestConnectionDestination = "Test_Connection_Destination",
+	CreateJobClicked = "Create_Job_Clicked",
+	CreateSourceClicked = "Create_Source_Clicked",
+	CreateDestinationClicked = "Create_Destination_Clicked",
+	AddCatalogClicked = "Add_Catalog_Clicked",
+	ImportCatalogFromDestinationClicked = "Import_Catalog_From_Destination_Clicked",
+	CatalogConnectClicked = "Catalog_Connect_Clicked",
```
deepanshupal09-datazip: wouldn't we also need an event for catalog connect success or failure event too? to track drop off from here or maybe a parameter inside catalog connect event itself, that contains catalog test connection status success or failure... similar to what we have in test connection source and destination

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
```diff
@@ -182,6 +202,19 @@ const ConfigureOptimizationModalView: React.FC<
...
+			option: getFrequencyOptionLabel(minorCron.frequency),
+		})
+		trackEvent(AnalyticsEvent.FrequencyOptionClicked, {
+			frequency: RUN_TYPE_LABEL.MEDIUM,
+			option: getFrequencyOptionLabel(majorCron.frequency),
+		})
+		trackEvent(AnalyticsEvent.FrequencyOptionClicked, {
+			frequency: RUN_TYPE_LABEL.FULL,
+			option: getFrequencyOptionLabel(fullCron.frequency),
+		})
```
deepanshupal09-datazip: here we fire 3 different events, one for each frequency option. this creates redundancy and makes the user's choice harder to track... 3 events means 3 separate rows in mixpanel, so we can't tell which combination a user actually selected, which is what the PRD needs.

instead we could send these configurations as properties on a single event (probably `ConfigurationSaveClicked`). that makes the user's combination easier to track and would give us better configuration distribution as well.

deepanshupal09-datazip: we can discuss this with product once before making any change.

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
```diff
@@ -80,9 +91,18 @@ const ScheduleSection: React.FC<ScheduleSectionProps> = ({
...
+						onChange={next => {
+							if (next !== "never") {
+								const tierEvent =
+									TIER_FREQUENCY_EVENT[
+										title as keyof typeof TIER_FREQUENCY_EVENT
+									]
+								if (tierEvent) {
+									trackEvent(tierEvent)
+								}
+							}
```
deepanshupal09-datazip: here, why do we fire this event everytime user changes selection? this would create unnecessary noise, ideally we should only fire event when user clicks save.

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
```diff
@@ -316,9 +353,15 @@ const ConfigureOptimizationModalView: React.FC<
...
-													)
+													const nextValue =
+														numericValue === "" ? 0 : Number(numericValue)
+													setTargetFileSize(nextValue)
+													if (nextValue !== DEFAULT_TARGET_FILE_SIZE) {
+														trackEvent(
+															AnalyticsEvent.TargetFileSizeCustomized,
+															{ target_file_size: nextValue },
+														)
+													}
```
deepanshupal09-datazip: this event is fired on every keystroke... assume user enter `512` ... so 3 events would be fired, `5`, `51` and `512`. we should only fire events when users clicks save.

server/internal/utils/telemetry/fusion.go
deepanshupal09-datazip, line 1: lets rename this to `optimization.go`, as this is the naming scheme we follow in BFF.

server/internal/utils/telemetry/fusion.go
```diff
@@ -0,0 +1,48 @@
...
+// waitForTelemetryInstance waits briefly for InitTelemetry's goroutine to set instance.
+func waitForTelemetryInstance() bool {
+	for i := 0; i < 50; i++ {
+		if instance != nil {
+			return true
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	return instance != nil
+}
```
deepanshupal09-datazip: instead of doing this, we can use channels, like this:

in `fusion.go`

```go
// startupEventTimeout bounds how long a startup event waits for InitTelemetry.
const startupEventTimeout = 5 * time.Second

// trackStartupEvent fires the event once InitTelemetry finishes, dropping it
// if telemetry never became available.
func trackStartupEvent(event string, properties map[string]interface{}) {
	go func() {
		select {
		case <-ready:
		case <-time.After(startupEventTimeout):
			return
		}
		if instance == nil {
			return
		}
		if err := TrackEvent(context.Background(), event, properties); err != nil {
			logger.Debugf("Failed to track %s event: %s", event, err)
		}
	}()
}

func TrackInstalledFusion() {
	trackStartupEvent(EventInstalledFusion, nil)
}

func TrackInstallationMode(mode string) {
	trackStartupEvent(EventInstallationMode, map[string]interface{}{"mode": mode})
}
```

and in `telemetry.go`

```go
var (
	instance *Telemetry
	// ready is closed once InitTelemetry's goroutine finishes (whether or not
	// telemetry was actually enabled), so waiters can safely read instance.
	ready = make(chan struct{})
)

func InitTelemetry(db *database.Database) {
	go func() {
		defer close(ready)

		if disabled, _ := strconv.ParseBool(os.Getenv("TELEMETRY_DISABLED")); disabled {
			return
		}
		// ... rest unchanged
	}()
}
```

server/internal/appconfig/appconfig.go
```diff
@@ -46,6 +46,7 @@ type Config struct {
 	OptimizationUsername           string
 	OptimizationPassword           string
 	OptimizationRequestTimeout     time.Duration
+	InstallationMode               string
```
deepanshupal09-datazip: this isn't really a config.. if you see other fields in this struct they are configuration that can be set via `env` or `app.yaml`, but `InstallationMode` is an automatically detected value.

since only telemetry uses it, better to do the `KUBERNETES_SERVICE_HOST` check inside `TrackInstallationMode()` itself and drop this field from config

ui/src/core/analytics/analyticsEnums.ts
```diff
@@ -1,7 +1,29 @@
...
+	StatusToggleOnClicked = "Status_Toggle_On_Clicked",
+	StatusToggleOffClicked = "Status_Toggle_Off_Clicked",
+	StatusEnableSuccessful = "Status_Enable_Successful",
+	StatusEnableFailed = "Status_Enable_Failed",
+	StatusRetryClicked = "Status_Retry_Clicked",
+	ViewLogsClicked = "View_Logs_Clicked",
+	HealthScoreTooltipViewed = "Health_Score_Tooltip_Viewed",
+	TargetFileSizeCustomized = "Target_File_size_Customized",
+	MaintenanceModuleOpened = "Maintenance_Module_Opened",
 }
```
deepanshupal09-datazip, line 29:
```suggestion
export enum AnalyticsEvent {
	// Test Connection
	TestConnectionSource = "test_connection_source",
	TestConnectionDestination = "test_connection_destination",

	// Creation flows
	CreateJobClicked = "create_job_clicked",
	CreateSourceClicked = "create_source_clicked",
	CreateDestinationClicked = "create_destination_clicked",

	// Catalog
	AddCatalogClicked = "add_catalog_clicked",
	ImportCatalogFromDestinationClicked = "import_catalog_from_destination_clicked",
	CatalogConnectClicked = "catalog_connect_clicked",

	// Metrics
	ViewTableMetricsClicked = "view_table_metrics_clicked",
	ViewMetricsClicked = "view_metrics_clicked",
	HealthScoreTooltipViewed = "health_score_tooltip_viewed",

	// Job configuration
	ConfigureButtonClicked = "configure_button_clicked",
	LiteFrequencySelected = "lite_frequency_selected",
	MediumFrequencySelected = "medium_frequency_selected",
	FullFrequencySelected = "full_frequency_selected",
	FrequencyOptionClicked = "frequency_option_clicked",
	TargetFileSizeCustomized = "target_file_size_customized",
	ConfigurationSaveClicked = "configuration_save_clicked",
	ConfigurationSaveSuccessful = "configuration_save_successful",
	ConfigurationSaveFailed = "configuration_save_failed",

	// Job status
	StatusToggleOnClicked = "status_toggle_on_clicked",
	StatusToggleOffClicked = "status_toggle_off_clicked",
	StatusEnableSuccessful = "status_enable_successful",
	StatusEnableFailed = "status_enable_failed",
	StatusRetryClicked = "status_retry_clicked",
	ViewLogsClicked = "view_logs_clicked",

	// Maintenance
	MaintenanceModuleOpened = "maintenance_module_opened",
}
```

lets keep the naming consistent, and keep it as `snake_case` as before.

ui/src/modules/maintenance/features/tables/pages/Tables.tsx
```diff
@@ -208,6 +215,11 @@ const Tables: React.FC = () => {
 				}
 				toggleTableOptimizing(request, {
 					onSuccess: result => {
+						if (enabled && result.success) {
+							trackEvent(AnalyticsEvent.StatusEnableSuccessful)
+						} else if (enabled && !result.success) {
+							trackEvent(AnalyticsEvent.StatusEnableFailed)
```
deepanshupal09-datazip: we also need to track failed on api failure... 
fix: add `onError` callback and emit enable failed event:

```
		onError: () => {
			if (request.enabled) trackEvent(AnalyticsEvent.StatusEnableFailed)
		},
```

ui/src/modules/maintenance/features/tables/pages/Tables.tsx
```diff
@@ -421,6 +435,7 @@ const Tables: React.FC = () => {
 				title="Failed to update optimization configuration"
 				error={optimizationErrorLogs.join("\n")}
 				onAction={() => {
+					trackEvent(AnalyticsEvent.StatusRetryClicked)
```
deepanshupal09-datazip: we emit retry clicked event here, but we dont track if retry actually succeeded or failed.

deepanshupal09-datazip (changes requested):
since we are adding a lot of new events here... every `trackEvent` call hits ipify + ipinfo to build system info, nothing is cached. with 28 events now, a single configure session can easily make a lot of these calls and ipinfo unauthenticated can get rate limited, after that location would degrade for all subsequent events.

can you check if we can fetch the ip/location once and cache it for the session instead of calling these external apis on every event?

ui/src/core/analytics/analyticsEnums.ts
deepanshupal09-datazip: can discuss this with product.

ui/src/core/layout/components/SidebarNavItem.tsx
ashi-droid: Yes, this is what is expected by product, they want the metrics per click and not per-user.

ui/src/core/analytics/analyticsEnums.ts
ashi-droid: added success or failure as a property in the same event

ui/src/core/layout/components/SidebarNavItem.tsx
ashi-droid: resolved by adding per session key for maintenance_module_opened

docker-compose-v1.yml
badalprasadsingh, line 271: We have to add this here, for `Fusion` to get the shared user-id file created by OLake-UI

```yaml
      - type: bind
        source: ${PWD}/olake-data/telemetry
        target: /tmp/olake-config/telemetry
        read_only: true
```

server/internal/handlers/handler.go
```diff
@@ -1,11 +1,13 @@
 package handlers
 
 import (
+	"os"
```
badalprasadsingh: Run `gofmt -v <path_of_file>`, **lint** is failing here

server/internal/utils/telemetry/constants.go
```diff
@@ -19,4 +19,6 @@ const (
 	EventDestinationCreated  = "destination_created"
 	EventSourcesUpdated      = "sources_updated"
 	EventDestinationsUpdated = "destinations_updated"
+	EventInstalledFusion     = "fusion_installed"
+	EventInstallationMode    = "installation_mode"
```
badalprasadsingh: Don't think this is being used anywhere.

server/internal/utils/telemetry/optimization.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/datazip-inc/olake-ui/server/internal/utils/logger"
+)
+
+// startupEventTimeout bounds how long a startup event waits for InitTelemetry.
+const startupEventTimeout = 5 * time.Second
```
badalprasadsingh: We have a timeout of `10mins` which we take from the `env`, reuse that here.

badalprasadsingh (changes requested):
Will check the Telemetry Metrics given by product after this review.

Please fix both the lint issues at `Frontend` and `Server`

server/internal/utils/telemetry/optimization.go
```diff
@@ -0,0 +1,39 @@
...
+	}()
+}
+
+func TrackInstalledFusion(serviceHost string) {
+	var mode string
+	if serviceHost != "" {
+		mode = "helm"
+	} else {
+		mode = "docker"
+	}
```
deepanshupal09-datazip:
```suggestion
	mode := utils.Ternary(serviceHost != "","helm", "docker")
```

server/main.go
```diff
@@ -63,12 +63,13 @@ func main() {
 
 	telemetry.InitTelemetry(db)
 
+	cfg := appconfig.Load()
+
```
deepanshupal09-datazip: this change is no longer required, lets revert this.

ui/src/core/analytics/analyticsService.ts
```diff
@@ -59,10 +73,28 @@ const getSystemInfo = async () => {
...
 
+const getSystemInfo = async (): Promise<SystemInfo> => {
+	try {
+		const cached = sessionStorage.getItem(SYSTEM_INFO_SESSION_KEY)
+		if (cached) {
+			return JSON.parse(cached) as SystemInfo
+		}
+	} catch (error) {
+		console.error("Error fetching system info:", error)
+	}
```
deepanshupal09-datazip: can we move this to a small sessionStore helper in `common/utils/session.ts` instead of using `sessionStorage` directly? 
this makes the logic reusable wherever we want to use it.
something like `sessionStore.get<boolean>(key)` / `sessionStore.set(key, true)`.

deepanshupal09-datazip: we do same in `SidebarNavItem.tsx`too

ui/src/core/layout/components/SidebarNavItem.tsx
```diff
@@ -1,6 +1,15 @@
 import clsx from "clsx"
 import { NavLink } from "react-router-dom"
 
+import { trackEvent, AnalyticsEvent } from "@/core/analytics"
+
+const MAINTENANCE_PATHS = new Set([
+	"/maintenance/tables",
+	"/maintenance/catalogs",
+])
+
```
deepanshupal09-datazip: hardcoding paths here means updating this file every time a maintenance item is added, and `SidebarNavItem` is generic (Settings uses it too) so it shouldn't know about a specific module. tracking on click also misses the collapsed sidebar, deep links and refreshes entirely.

since `nav.ts` already has the `paths` and `matchesPath` already maps a `pathname` to a module (breadcrumbs use it), we can detect this from the route instead — a small hook in `core/layout/hooks`:

```typescript
export const useModuleAnalytics = (modules: NavModule[]) => {
	const { pathname } = useLocation()

	useEffect(() => {
		const mod = modules.find(m =>
			m.items.some(i => matchesPath(pathname, i.path)),
		)
		const event = mod && MODULE_OPENED_EVENTS[mod.key]
		if (!event) return

		const key = `analytics_fired_${event}`
		if (sessionStore.get<boolean>(key)) return

		trackEvent(event, { path: pathname })
		sessionStore.set(key, true)
	}, [pathname, modules])
}
```
called once from `layout/index.tsx` (it already has `navModules` and `location`), with `MODULE_OPENED_EVENTS` being a small `{ maintenance: AnalyticsEvent.MaintenanceModuleOpened }` map in the hook file. then this file drops the paths, the session key and the analytics import entirely,

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -188,8 +190,14 @@ const CatalogModal: React.FC<CatalogModalProps> = ({
 				)
 				setCreatedCatalogName(getCatalogNameFromFormData(formData))
 			}
+			trackEvent(AnalyticsEvent.CatalogConnectClicked, {
+				status: "success",
+			})
```
deepanshupal09-datazip, line 195:
```suggestion
			trackEvent(AnalyticsEvent.CatalogConnectClicked, { success: true })
```
lets make this boolean.
just to keep it consistent with previous test connection event.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -188,8 +190,14 @@ const CatalogModal: React.FC<CatalogModalProps> = ({
...
 			setActiveModal(ActiveCatalogModalState.CATALOG_SUCCESS)
 		} catch (e) {
+			trackEvent(AnalyticsEvent.CatalogConnectClicked, {
+				status: "failed",
+			})
```
deepanshupal09-datazip, line 200:
```suggestion
			trackEvent(AnalyticsEvent.CatalogConnectClicked, { success: false })
```
same here.

ui/src/modules/maintenance/features/tables/pages/Tables.tsx
```diff
@@ -422,7 +439,24 @@ const Tables: React.FC = () => {
 				error={optimizationErrorLogs.join("\n")}
 				onAction={() => {
 					setOptimizationErrorOpen(false)
-					if (lastToggleRequest) toggleTableOptimizing(lastToggleRequest)
+					if (lastToggleRequest) {
+						toggleTableOptimizing(lastToggleRequest, {
+							onSuccess: result => {
+								trackEvent(AnalyticsEvent.StatusRetryClicked, {
+									status: result.success ? "success" : "failed",
+								})
```
deepanshupal09-datazip, line 448:
```suggestion
								trackEvent(AnalyticsEvent.StatusRetryClicked, {
									success: result.success,
								})
```

ui/src/modules/maintenance/features/tables/pages/Tables.tsx
```diff
@@ -422,7 +439,24 @@ const Tables: React.FC = () => {
...
+								if (!result.success) {
+									setOptimizationErrorLogs(result.logs ?? [])
+									setOptimizationErrorOpen(true)
+								}
+							},
+							onError: () => {
+								trackEvent(AnalyticsEvent.StatusRetryClicked, {
+									status: "failed",
+								})
```
deepanshupal09-datazip, line 457:
```suggestion
								trackEvent(AnalyticsEvent.StatusRetryClicked, {
									success: false,
								})
```

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
```diff
@@ -316,9 +333,9 @@ const ConfigureOptimizationModalView: React.FC<
 														/[^0-9]/g,
 														"",
 													)
-													setTargetFileSize(
-														numericValue === "" ? 0 : Number(numericValue),
-													)
+													const nextValue =
+														numericValue === "" ? 0 : Number(numericValue)
+													setTargetFileSize(nextValue)
```
deepanshupal09-datazip: this change is not needed.

ui/src/core/layout/components/SidebarNavItem.tsx
```diff
@@ -10,6 +19,17 @@ const SidebarNavItem: React.FC<{
 }> = ({ path, label, icon: Icon, iconSize = 14, className }) => (
 	<NavLink
 		to={path}
+		onClick={() => {
+			if (
+				MAINTENANCE_PATHS.has(path) &&
+				!sessionStorage.getItem(MAINTENANCE_SESSION_KEY)
+			) {
+				console.log("Maintenance module opened", path)
```
deepanshupal09-datazip: remove this console log.

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
```diff
@@ -80,9 +85,9 @@ const ScheduleSection: React.FC<ScheduleSectionProps> = ({
 					</label>
 					<Select
 						value={value.frequency}
-						onChange={next =>
+						onChange={next => {
 							onChange({ ...value, frequency: next, customCron: "" })
-						}
+						}}
```
deepanshupal09-datazip: this change is not needed.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
deepanshupal09-datazip, line 175: do we need to send event here also? 
when user clicks on connect but form validation failed.

deepanshupal09-datazip:
also please fix the linter errors.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
deepanshupal09-datazip: not needed.

ui/src/core/analytics/analyticsEnums.ts
```diff
@@ -1,7 +1,36 @@
...
+	FullFrequencySelected = "full_frequency_selected",
+	FrequencyOptionClicked = "frequency_option_clicked",
+	TargetFileSizeCustomized = "target_file_size_customized",
+	ConfigurationSaveClicked = "configuration_save_clicked",
+	ConfigurationSaveSuccessful = "configuration_save_successful",
+	ConfigurationSaveFailed = "configuration_save_failed",
+	// Job status
+	StatusToggleOnClicked = "status_toggle_on_clicked",
+	StatusToggleOffClicked = "status_toggle_off_clicked",
+	StatusEnableSuccessful = "status_enable_successful",
```
deepanshupal09-datazip: here also instead of firing three events, we could fire only one event `StatusToggleOnClicked` and pass `success` parameter inside it

server/internal/utils/telemetry/constants.go
ashi-droid: EventInstalledFusion is used, removed EventInstallationMode

ui/pnpm-workspace.yaml
deepanshupal09-datazip, line 1: this file is not needed, we already have overrides in package.json

ui/package.json
deepanshupal09-datazip, line 1: many of the dependency upgrades in this file aren't needed. only `fast-uri` and `antd` one are needed.

deepanshupal09-datazip: also prettier version was upgraded which is causing lint changes in many files.
please revert all the dependency upgrades except fast-uri and antd. easiest way is 
`git checkout staging -- ui/package.json ui/pnpm-lock.yaml` (make sure you have the latest staging code) , then re-apply just those two and run `pnpm install` to regenerate the lockfile

server/internal/utils/telemetry/constants.go
```diff
@@ -19,4 +19,5 @@ const (
 	EventDestinationCreated  = "destination_created"
 	EventSourcesUpdated      = "sources_updated"
 	EventDestinationsUpdated = "destinations_updated"
+	EventInstalledFusion     = "installed_fusion"
 )
```
deepanshupal09-datazip: InstallationMode removed from here?

server/internal/handlers/handler.go
```diff
@@ -28,6 +31,8 @@ func NewHandler(appSvc *services.AppService, cfg *appconfig.Config, db *database
 
 	if opt := appSvc.Optimization(); opt != nil {
 		handler.Optimization = optimization.NewHandler(opt)
+		KubernetesServiceHost := os.Getenv("KUBERNETES_SERVICE_HOST")
```
deepanshupal09-datazip:
```suggestion
		kubernetesServiceHost := os.Getenv("KUBERNETES_SERVICE_HOST")
```

ui/src/modules/maintenance/features/tables/pages/Tables.tsx
```diff
@@ -200,6 +202,11 @@ const Tables: React.FC = () => {
 				})
 			},
 			onToggleOptimizingStatus: (row, enabled) => {
+				trackEvent(
+					enabled
+						? AnalyticsEvent.StatusToggleOnClicked
+						: AnalyticsEvent.StatusToggleOffClicked,
+				)
```
deepanshupal09-datazip: should send event only when API call is finished, else duplicate events will be there... one with success parameter and one without it.

ui/src/common/utils/session.ts
```diff
@@ -0,0 +1,25 @@
...
+	try {
+		sessionStorage.setItem(key, JSON.stringify(value))
+	} catch (error) {
+		console.error(`sessionStore: failed to write "${key}"`, error)
+	}
+}
+
+const sessionStore = { get, set }
+
+export default sessionStore
```
deepanshupal09-datazip: minor: make this a named export and add it to `index.ts` of this folder. and then use directly like:
`import { sessionStore } from "@/common/utils"`

server/internal/utils/telemetry/optimization.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"time"
+
+	"github.com/datazip-inc/olake-ui/server/internal/appconfig"
+	"github.com/datazip-inc/olake-ui/server/internal/utils"
+	"github.com/datazip-inc/olake-ui/server/internal/utils/logger"
+)
+
+// startupEventTimeout bounds how long a startup event waits for InitTelemetry.
+var startupEventTimeout = appconfig.Load().OptimizationRequestTimeout
```
deepanshupal09-datazip: this should not be `OptimizationRequestTimeout`, we do not need to wait 10min for telemetry.
I think 10 seconds should be enough for this.

ui/src/core/analytics/analyticsEnums.ts
```diff
@@ -1,7 +1,34 @@
...
+	ViewTableMetricsClicked = "view_table_metrics_clicked",
+	ViewMetricsClicked = "view_metrics_clicked",
+	HealthScoreTooltipViewed = "health_score_tooltip_viewed",
+	// Job configuration
+	ConfigureButtonClicked = "configure_button_clicked",
+	LiteFrequencySelected = "lite_frequency_selected",
+	MediumFrequencySelected = "medium_frequency_selected",
+	FullFrequencySelected = "full_frequency_selected",
+	FrequencyOptionClicked = "frequency_option_clicked",
+	TargetFileSizeCustomized = "target_file_size_customized",
```
deepanshupal09-datazip: these are not being used anywhere now.

server/internal/utils/telemetry/constants.go
deepanshupal09-datazip: stlll pending.

server/internal/utils/telemetry/telemetry.go
deepanshupal09-datazip:
```suggestion
	client := http.Client{Timeout: TelemetryConfigTimeout}
	resp, err := client.Get(IPUrl)
```

server/internal/utils/telemetry/constants.go
ashi-droid: yes, installation mode is sent as a parameter to FusionInstalled

defined in 

> `server/internal/utils/telemetry/optimization.go`

server/internal/utils/telemetry/optimization.go
ashi-droid: @badalprasadsingh

badalprasadsingh: @deepanshupal09-datazip, we have made the timeout constant to `10min` across all APIs in optimization part.

So let's keep it here like this as well.

deepanshupal09-datazip: this is not an optimization api. this is telemetry initialization timeout, it is not related to optimization apis.

badalprasadsingh: I discussed this with Ankit, lets keep it 10secs.

deepanshupal09-datazip (dismissed):
lgtm

server/internal/utils/telemetry/telemetry.go
```diff
@@ -46,6 +48,8 @@ type Telemetry struct {
 }
 
 func InitTelemetry(db *database.Database) {
+	enableOptimization := appconfig.Load().EnableOptimization
```
deepanshupal09-datazip: can we pass this `enableOptimization` through function parameter in `main.go` like we do for `db`?
this makes the dependency explicit.

badalprasadsingh: It's non-blocking for now. But, @ashi-droid a very clean fix for this would be:

To move `cfg := appconfig.Load()` at the very beginning under `logger.Init()` in main function, and pass `cfg.EnableOptimization` as a param to both `InitAppService` and `InitTelemetry`. This way making both the init functions `InitAppService` and `InitTelemetry` take enable optimization as a param.

deepanshupal09-datazip (dismissed):

server/internal/utils/telemetry/telemetry.go
```diff
@@ -227,3 +231,10 @@ func GetVersion() string {
 	}
 	return ""
 }
+
+func trackFusionInstalled() {
+	mode := utils.Ternary(os.Getenv("KUBERNETES_SERVICE_HOST") != "", "helm", "docker")
```
hash-data: currently we not have helm or docker installation separation in normal ui deployment can we add that as well

ui/src/core/analytics/analyticsEnums.ts
```diff
@@ -1,7 +1,29 @@
...
+	HealthScoreTooltipViewed = "health_score_tooltip_viewed",
+	// Job configuration
+	ConfigureButtonClicked = "configure_button_clicked",
+	ConfigurationSaveClicked = "configuration_save_clicked",
+	ConfigurationSaveSuccessful = "configuration_save_successful",
+	ConfigurationSaveFailed = "configuration_save_failed",
+	// Job status
+	StatusToggleOnClicked = "status_toggle_on_clicked",
+	StatusToggleOffClicked = "status_toggle_off_clicked",
+	StatusRetryClicked = "status_retry_clicked",
```
hash-data: did not get this

docker-compose-v1.yml
```diff
@@ -269,6 +269,10 @@ services:
       - kind-kubeconfig:/root/.kube:ro
       - spark-home:/opt/spark:ro
       - <<: *amoroLogVolumeDetails
+      - type: bind
+        source: ${PWD}/olake-data/telemetry
+        target: /tmp/olake-config/telemetry
```
hash-data, line 274: any reason?

server/main.go
```diff
@@ -46,29 +46,31 @@ import (
...
-	appSvc, err := services.InitAppService(db)
+	appSvc, err := services.InitAppService(db, enableOptimization)
 	if err != nil {
 		logger.Fatalf("Failed to initialize services: %s", err)
 		return
 	}
 	logger.Info("Application services initialized successfully")
 
-	telemetry.InitTelemetry(db)
+	telemetry.InitTelemetry(db, enableOptimization)
```
hash-data: why we need to send it in init telemetry?

server/main.go
badalprasadsingh, line 1: Revert back these changes in `main.go` no longer required.

server/internal/services/services.go
badalprasadsingh, line 1: Revert back these changes in this `services.go` no longer required.

server/internal/services/optimization/terminal.go
```diff
@@ -47,6 +48,38 @@ func createAlterQuery(database, table string, properties map[string]string) stri
...
 	tables := config.Tables
 	properties := convertConfigToMap(config)
+	bulkConfigured := len(tables) >= 2
+
+	// Preload metrics + "updated" before ALTER (optional but needed for the contract).
+	type tableTelemetry struct {
+		name      string
+		tableSize string
+		fileCount float64
+		updated   bool
```
badalprasadsingh: So, updated is no longer required. (Have conveyed to Siddharth).

Reason: We always "update" on saving a cron, we never get any default value. So this logic goes unnecessary.

server/internal/services/optimization/terminal.go
```diff
@@ -47,6 +48,38 @@ func createAlterQuery(database, table string, properties map[string]string) stri
...
+		t := tableTelemetry{name: tableName}
+		if details, err := s.getTableDetails(ctx, catalog, database, tableName); err == nil {
+			if detailsMap, ok := details.(map[string]interface{}); ok {
+				if bm, ok := detailsMap["baseMetrics"].(map[string]interface{}); ok {
+					if size, ok := bm["totalSize"].(string); ok {
+						t.tableSize = size
+					}
+					if fc, ok := bm["fileCount"].(float64); ok {
+						t.fileCount = fc
+					}
```
badalprasadsingh: Have we tested "filecount"?

server/internal/services/optimization/terminal.go
```diff
@@ -47,6 +48,38 @@ func createAlterQuery(database, table string, properties map[string]string) stri
...
+				}
+				if props, ok := detailsMap["properties"].(map[string]interface{}); ok {
+					_, hasMinor := props[constants.OptMinorCron]
+					_, hasMajor := props[constants.OptMajorCron]
+					_, hasFull := props[constants.OptFullCron]
+					t.updated = hasMinor || hasMajor || hasFull
+				}
```
badalprasadsingh: This is logically wrong, UI always sends the cron to server. Anyways this should get removed, as we no longer require `updated`

server/internal/services/optimization/terminal.go
```diff
@@ -47,6 +48,38 @@ func createAlterQuery(database, table string, properties map[string]string) stri
 func (s *Service) SetProperties(ctx context.Context, catalog, database string, config dto.OptimizationTableConfig) (*dto.TableProperties, error) {
 	tables := config.Tables
 	properties := convertConfigToMap(config)
+	bulkConfigured := len(tables) >= 2
+
+	// Preload metrics + "updated" before ALTER (optional but needed for the contract).
+	type tableTelemetry struct {
+		name      string
```
badalprasadsingh: why do we need `name` for tables?

ui/src/modules/maintenance/features/tables/pages/RunHistory.tsx
```diff
@@ -128,22 +129,44 @@ const RunHistory: React.FC = () => {
...
+		catalog: decodedCatalog,
+		database: decodedDatabase,
+		table_name: decodedTableName,
+		table_size: tableDetails?.totalSize ?? "--",
+		run_id: row.runId,
+		status: row.status,
+		type: row.type,
+		start_time: row.startTime,
+		duration: row.duration,
+	})
```
badalprasadsingh: Are we sending all these on view logs and view compaction run metrics? I don't think we need so much.

I don't think we are allowed to send: `catalog`, `database`, `table_name`, `run_id`, `start_time`.

Confirm if this is even required in PRD. Correct me if we are not sending it Telemetry API.

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalView.tsx
badalprasadsingh, line 1: Non-blocking: unnecessary formatting

ui/src/modules/maintenance/features/catalogs/pages/Catalogs.tsx
```diff
@@ -32,6 +33,12 @@ const Catalogs: React.FC = () => {
 		setModalOpen(false)
 		setActiveCatalogName(undefined)
 	}
+
+	const openNewCatalogModal = () => {
+		trackEvent(AnalyticsEvent.AddCatalogClicked)
+		setModalOpen(true)
+	}
```
badalprasadsingh, line 40: Catalog Added Event will move to Fusion. Both failed or success. Remove all the logic for it from UI.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -39,6 +40,16 @@ const getCatalogNameFromFormData = (data: CatalogFormData): string => {
...
 
+const getCatalogTypeFromFormData = (data: CatalogFormData): string => {
+	const writer = (data as { writer?: { catalog_type?: string } }).writer
+	return writer?.catalog_type ?? "unknown"
+}
+
+const getOlakeImportedFromFormData = (data: CatalogFormData): boolean => {
+	const writer = (data as { writer?: { olake_imported?: boolean } }).writer
+	return !!writer?.olake_imported
+}
```
badalprasadsingh: For every catalog that is imported from OLake, we store it as a property in catalog properties in Fusion: `olake_created`. Use that while you move the logic to Fusion.

Thus, these functions will get removed from UI side. Remove all such unnecessary functions.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -187,9 +203,28 @@ const CatalogModal: React.FC<CatalogModalProps> = ({
...
 			}
+			trackEvent(AnalyticsEvent.CatalogConnectClicked)
 			setActiveModal(ActiveCatalogModalState.CATALOG_SUCCESS)
 		} catch (e) {
+			if (isEditMode) {
+				trackEvent(AnalyticsEvent.CatalogUpdated, {
+					catalog_type: getCatalogTypeFromFormData(formData),
+					olake_imported: getOlakeImportedFromFormData(formData),
+					status: "failed",
+				})
```
badalprasadsingh: All catalog events will move to Fusion: even editing it

ui/src/core/analytics/analyticsService.ts
```diff
@@ -49,7 +51,21 @@ const getLocationInfo = async (ip: string) => {
...
+	device_cpu: string
+	ip_address: string
+	location:
+		| {
+				country: string
+				region: string
+				city: string
+		  }
+		| ""
+}
```
badalprasadsingh, line 66: Why do we need this? Did @deepanshupal09-datazip approve of this?

As we are already sending this in Fusion, I don't think we require it in UI.

badalprasadsingh (changes requested):
Few Major.

ui/src/modules/maintenance/features/catalogs/pages/Catalogs.tsx
ashisharma-dz: Moved catalog created and updated to Fusion, but this is a click event. AddCatalogClicked. This should remain on the UI side.

badalprasadsingh: All click events will remain on UI side.

Any logic other than that: like Catalog Creation Event, Catalog Update Fail event, etc. will stay on backend (Fusion) side.

ui/src/core/analytics/analyticsService.ts
deepanshupal09-datazip: some events are UI only, and this `SystemInfo` data is required for them.

ui/src/modules/maintenance/features/tables/pages/RunHistory.tsx
deepanshupal09-datazip: aren't `catalog`, `database`, `table_name` sensitive information, I don't think we should be sending these to telemetry. Telemetry should only track anonymous user statistics. Once confirm this with product.

server/internal/services/optimization/terminal.go
ashisharma-dz: yes

server/internal/utils/telemetry/telemetry.go
```diff
@@ -232,9 +227,13 @@ func GetVersion() string {
...
-	}
+func TrackConfigurationSaved(ctx context.Context, properties map[string]interface{}) {
+	go func() {
+		if instance == nil {
+			return
+		}
+		if err := TrackEvent(ctx, EventConfigurationSaved, properties); err != nil {
+			logger.Debug("Failed to track configuration saved event: %s", err)
+		}
+	}()
```
deepanshupal09-datazip: lets move this out to `telemetry/optimization.go` like before, this file should have only init code.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -40,6 +40,16 @@ const getCatalogNameFromFormData = (data: CatalogFormData): string => {
...
 
+const getCatalogTypeFromFormData = (data: CatalogFormData): string => {
+	const writer = (data as { writer?: { catalog_type?: string } }).writer
+	return writer?.catalog_type ?? "unknown"
+}
+
+const getOlakeImportedFromFormData = (data: CatalogFormData): boolean => {
+	const writer = (data as { writer?: { olake_imported?: boolean } }).writer
+	return !!writer?.olake_imported
+}
```
deepanshupal09-datazip: these are not being used anywhere, please remove this.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
```diff
@@ -190,14 +199,9 @@ const CatalogModal: React.FC<CatalogModalProps> = ({
 				)
 				setCreatedCatalogName(getCatalogNameFromFormData(formData))
 			}
-			trackEvent(AnalyticsEvent.CatalogConnectClicked, {
-				success: true,
-			})
+			trackEvent(AnalyticsEvent.CatalogConnectClicked)
```
deepanshupal09-datazip: we should move this to above the `try` block, right now this won't be fired if api fails.

ui/src/modules/maintenance/features/tables/pages/RunHistory.tsx
```diff
@@ -6,15 +6,16 @@ import { useNavigate, useParams } from "react-router-dom"
...
 import { RunMetricsSidebar } from "../components"
-import { useRunHistoryFilters, useTableRuns } from "../hooks"
+import { useRunHistoryFilters, useTableDetails, useTableRuns } from "../hooks"
 import type { RunHistoryFilter, RunMetricRow, TableRun } from "../types"
 import { getRunLogsStatusConfig } from "../utils"
 
 const getColumns = (
 	handleMetricsClick: (row: TableRun) => void,
-	handleLogsClick: (runId: string) => void,
+	handleLogsClick: (runId: TableRun) => void,
```
deepanshupal09-datazip:
```suggestion
	handleLogsClick: (run: TableRun) => void,
```
not an `id` now.

ui/src/modules/maintenance/features/tables/pages/RunHistory.tsx
```diff
@@ -128,22 +129,39 @@ const RunHistory: React.FC = () => {
...
 	const totalRuns = data?.total ?? 0
 
 	const totalPages = Math.max(1, Math.ceil(totalRuns / pageSize))
 
+	const getRunEventProperties = (row: TableRun) => ({
+		table_size: tableDetails?.totalSize ?? "--",
+		status: row.status,
+		type: row.type,
+		duration: row.duration,
+	})
```
deepanshupal09-datazip:
```suggestion
	const getRunEventProperties = (run: TableRun) => ({
		table_size: tableDetails?.totalSize ?? "--",
		status: run.status,
		type: run.type,
		duration: run.duration,
	})
```

ui/src/modules/maintenance/features/tables/pages/RunLogs.tsx
```diff
@@ -1,14 +1,19 @@
...
+
 import { RunLogSidebar, RunLogsPanel } from "../components"
 import { DRIVER_SOURCE_KEY } from "../constants"
 import { tableKeys } from "../constants/queryKeys"
 
 const RunLogs: React.FC = () => {
 	const navigate = useNavigate()
+	useEffect(() => {
+		trackEvent(AnalyticsEvent.ViewRunsAndLogsClicked)
+	}, [])
```
deepanshupal09-datazip: This event should be fired when user clicks on "View Logs & Runs" on the Tables Page. This is the logs page.

Add this:

```
onViewLogs: row => {
	trackEvent(AnalyticsEvent.ViewRunsAndLogsClicked)
	navigate(getTableRunsPath(row.name))
},
```
in `Tables.tsx`

server/internal/services/optimization/terminal.go
```diff
@@ -68,6 +93,37 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database string, c
...
+				"table_size":         t.tableSize,
+				"file_count":         t.fileCount,
+				"lite_frequency":     properties[constants.OptMinorCron],
+				"medium_frequency":   properties[constants.OptMajorCron],
+				"full_frequency":     properties[constants.OptFullCron],
+				"target_file_size":   properties[constants.OptTargetFileSize],
+				"status":             status,
+				"bulk_configured":    bulkConfigured,
+				"compaction_enabled": compactionEnabled,
+				"updated":            t.updated,
```
deepanshupal09-datazip: please removed `updated` from here too.

server/internal/utils/telemetry/telemetry.go
```diff
@@ -227,3 +226,14 @@ func GetVersion() string {
...
 	return ""
 }
+
+func TrackConfigurationSaved(ctx context.Context, properties map[string]interface{}) {
+	go func() {
+		if instance == nil {
+			return
+		}
+		if err := TrackEvent(ctx, EventConfigurationSaved, properties); err != nil {
+			logger.Debug("Failed to track configuration saved event: %s", err)
```
deepanshupal09-datazip:
```suggestion
			logger.Debugf("Failed to track configuration saved event: %s", err)
```

server/internal/services/optimization/terminal.go
```diff
@@ -68,6 +93,37 @@ func (s *Service) SetProperties(ctx context.Context, catalog, database string, c
...
+			}
+			telemetry.TrackConfigurationSaved(ctx, props)
```
deepanshupal09-datazip: we are sending event for each table in case of bulk configure, is this expected?

server/internal/services/optimization/terminal.go
```diff
@@ -47,6 +48,30 @@ func createAlterQuery(database, table string, properties map[string]string) stri
...
 	properties := convertConfigToMap(config)
+	bulkConfigured := len(tables) >= 2
+
+	// Preload metrics + "updated" before ALTER (optional but needed for the contract).
+	type tableTelemetry struct {
+		tableSize string
+		fileCount float64
+	}
+	pre := make([]tableTelemetry, 0, len(tables))
+	for _, tableName := range tables {
```
deepanshupal09-datazip: this loop does N api call, here also we should add check like below before doing api calls:

```
	if config.SQLInput.MinorCron != nil ||
		config.SQLInput.MajorCron != nil ||
		config.SQLInput.FullCron != nil {
```

deepanshupal09-datazip:
Also, please fix all the bulid and lint errors.

server/internal/utils/telemetry/telemetry.go
badalprasadsingh: `ConfigureButtonClicked` should have a single event which will be in UI side. So, this will get removed altogether.

ui/src/modules/maintenance/features/catalogs/components/CatalogModal.tsx
deepanshupal09-datazip: still pending.

ui/src/core/layout/hooks/useModuleAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+		const mod = modules.find(m =>
+			m.items.some(i => matchesPath(pathname, i.path)),
+		)
+		const event = mod && MODULE_OPENED_EVENTS[mod.key]
+		if (!event) return
+
+		const key = `analytics_fired_${event}`
+		if (sessionStore.get<boolean>(key)) return
+
+		trackEvent(event, { path: pathname })
```
deepanshupal09-datazip:
```suggestion
		trackEvent(event)
```
we should remove `path` from here, as it might contain table name or database name

ui/src/modules/maintenance/features/tables/components/modals/ConfigureOptimizationModalBulk.tsx
```diff
@@ -115,6 +115,10 @@ const ConfigureOptimizationModalBulk: React.FC<
 					: undefined
 			}
 			isSaving={isSaving}
+			telemetry={{
+				bulkConfigured: true,
+				tableCount: tables.length,
+			}}
```
deepanshupal09-datazip: we should not be passing `telemetry` as a prop, telemetry is a side-concern and should be handled separately, This component is purely presentational and should not handle telemetry.

Instead we could make a utils like we did for `trackTestConnection` in `analyticsUtils`:
```
type ConfigurationSavedContext = { ... }

export const trackConfigurationSaved = (
	payload: UpdateTableCronApiRequest,
	context: ConfigurationSavedContext,
) => { ... }
```
and call it from the mutate callbacks in both Single and Bulk:
```
const telemetryContext = { bulkConfigured: true, tableCount: tables.length }

onSuccess: result => {
	trackConfigurationSaved(payload, {
		...telemetryContext,
		status: result.success ? "success" : "failed",
	})
	if (result.success) {
		onSuccess()
		return
	}
	onError(result.logs ?? (result.message ? [result.message] : []))
},
onError: () =>
	trackConfigurationSaved(payload, { ...telemetryContext, status: "failed" }),
```

ui/src/core/analytics/analyticsService.ts
```diff
@@ -59,10 +75,20 @@ const getSystemInfo = async () => {
...
 }
 
+const getSystemInfo = async (): Promise<SystemInfo> => {
+	const cached = sessionStore.get<SystemInfo>(SYSTEM_INFO_SESSION_KEY)
+	if (cached) {
+		return cached
+	}
+
+	const systemInfo = await fetchSystemInfo()
+	sessionStore.set(SYSTEM_INFO_SESSION_KEY, systemInfo)
```
deepanshupal09-datazip:
```suggestion
    if (systemInfo.ip_address && systemInfo.location) {
    	sessionStore.set(SYSTEM_INFO_SESSION_KEY, systemInfo)
    }
```

deepanshupal09-datazip (approved):
LGTM

PR #412, opened 2026-08-10
fix(opt): compile fix in config yaml

badalprasadsingh:
# Description

The file `config.yaml` had a compile error with a dangling variable.


badalprasadsingh:
Closing this PR as this is required right now, and removing it will causes multiple backward compatibility issues.

PR #413, opened 2026-08-10
fix(opt): update spec accoring to new destination revamp

badalprasadsingh:
# Description

Updating `spec.json` according to the new destination configuration revamp in `OLake-Go`.  

Changes in the Optimization Spec, which is "intentionally" kept different from OLake-Go `spec` logic:
1. Remove `arrow_writes`
2. Keep `catalog_name` required
3. Remove the pattern regex in `catalog_name`
4. Add `catalog_name` in Glue Catalog Type as it is not there in the new destination spec in OLake-Go
5. [Already there] `rest_signing_v_4` set to value `false` as default

## Type of change

# How Has This Been Tested?

- [x] `OLake-UI` with `OLake-Fusion` : `Polaris`


server/internal/models/catalog_config.go
```diff
@@ -17,6 +17,7 @@ const (
 	RestCatalog CatalogType = "rest"
 )
 
+// TODO: Test BigLake with Optimization Service or Suppress it until further testing
```
hash-data: move this as issue in olake fusion

server/internal/services/optimization/mapper.go
```diff
@@ -145,15 +150,37 @@ func mapAuthConfig(olakeConfig *models.Config, authConfig, cmpStorageConfig map[
...
+
+	// Below logic is similar to what we have in OLake-Go : destination/iceberg/config.go
+	// S3 tables use SigV4 authentication and require signing name to be set to "s3tables"
+	if olakeConfig.CatalogType == "s3tables" {
+		olakeConfig.RestSigningV4 = true
+		olakeConfig.RestSigningName = "s3tables"
+	}
+	// Unity Catalog doesn't support identifier fields (disable them)
+	if olakeConfig.CatalogType == "unity" {
+		olakeConfig.NoIdentifierFields = true
```
hash-data: we dont need to map this property

server/internal/services/optimization/resources/spec.json
```diff
@@ -11,11 +11,36 @@
             "oneOf": [
               {
                 "properties": {
+                  "aws_access_key": {
```
hash-data, line 14: recetly there was a fix on olake cli for auth can you check that once?

server/internal/constants/constants.go
```diff
@@ -119,18 +119,29 @@ var (
...
+	OptOLakeCatalogType = "olake-catalog-type"
 	// OptimizeTableFormatList defines supported table formats for catalogs
 	OptimizeTableFormatList = []string{"ICEBERG"}
 	// hard-coding to S3 now, as the other options are "hadoop" & "OSS" for optimization
 	// GCS & ADLS are supported, given the catalog manages the sdk (eg, Lakekeeper with GCS flavour)
 	DefaultOptimizationStorageType = "S3"
-	CatalogSpecVersion             = "v0.7.4"
+	CatalogSpecVersion             = "v0.9.1"
 	CatalogSpecType                = "iceberg"
 	IcebergCatalogSpecFile         = "server/internal/services/optimization/resources/spec.json"
```
hash-data, line 134: do we need to maintain this or can we think of something else?

badalprasadsingh: We are maintaining this : to track last CLI `spec.json` it has been aligned with. 

The only other option is to make `spec.json` in UI a complete `response.json` and keep it there as field.

server/internal/services/optimization/resources/spec.json
badalprasadsingh: The last spec change in CLI was in this [PR](https://github.com/datazip-inc/olake/pull/1087), and this is absolutely aligned with it.

server/internal/services/optimization/mapper.go
badalprasadsingh: Removed it, but this comes out to be a change which is now different from what we have in olake-go

server/internal/models/catalog_config.go
badalprasadsingh: Removed

server/internal/services/optimization/resources/spec.json
badalprasadsingh: Added the changes from the discussed PR.

hash-data (approved):

PR #415, opened 2026-08-11
fix: telemetry events

deepanshupal09-datazip:
# Description

The worker's sync-telemetry callback now carries a pre-resolved `properties` dict that olake-ui forwards as-is; the old DB-enrichment path stays only as a fallback for un-upgraded workers. Also fixes sync event bugs and adds a community-help banner on sync failures.

- Forwarded events get `catalog` filled in from the DB (the one thing only the CLI would otherwise know, and these events mean the CLI wasn't emitting).
- Fixed: `started_at` was never emitted (every event carried `ended_at`); a missing `stats.json` or a decryption failure no longer drops the event.
- New: `sync_cancelled` event; `source_version`/`destination_version`/`sync_run_count`/`catalog` properties on enriched events.
- Frontend: `CommunityHelpBanner` on Job History when the latest run failed (links to olake.io/slack, session-dismissible), plus `community_help_clicked` and `job_paused` analytics events.

## Type of change

# How Has This Been Tested?

- [x] Callback with `properties` forwards verbatim; without it, DB enrichment used — both verified in Mixpanel
- [x] `started_at`/`ended_at` on correct events; `completed` delivered even with `stats.json` missing
- [x] `cancelled` callback → `sync_cancelled`; pause fires `job_paused`, resume doesn't
- [x] Banner show/hide/dismiss works; CTA fires `community_help_clicked`

# Screenshots or Recordings

## Related PR's (If Any):
- datazip-inc/olake-helm#188


server/internal/utils/telemetry/sync.go
```diff
@@ -13,28 +13,50 @@ import (
...
-		if job == nil {
-			return nil, fmt.Errorf("job not found")
+	job, err := instance.db.GetJobByID(jobID, true)
+	if err != nil {
+		// A decryption failure must not cost us the whole event - retry
+		// without decrypt; catalog just falls back to "none" below.
+		job, err = instance.db.GetJobByID(jobID, false)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get job details: %s", err)
 		}
```
vishalm0509: eralier even DB failure runs a retry. We can do this instead. 
```suggestion
	job, err := instance.db.GetJobByID(jobID, true)
	if errors.Is(err, constants.ErrConfigDecrypt) {
		// Decryption failure must not cost us the whole event - retry
		// without decrypt; catalog falls back to "unknown" below.
		job, err = instance.db.GetJobByID(jobID, false)
	}
	if err != nil {
		return nil, fmt.Errorf("failed to get job details: %s", err)
```

server/internal/utils/telemetry/sync.go
```diff
@@ -130,79 +184,76 @@ func addStatsProperties(properties map[string]interface{}, mainSyncDir string) e
...
+			selected[ns+"."+s.StreamName] = true
+			selectedCount++
+			if s.Normalization {
 				normalizedCount++
 			}
-			if stream.PartitionRegex != "" {
+			if s.PartitionRegex != "" {
 				partitionedCount++
 			}
 		}
```
vishalm0509, line 248: CLI is the source of truth for these events - let's have it in the CLI only. We can have a discussuion with tech and product regd. this.  

We will not define the type of structs defined in the CLI here - or else we will have to make changes here as well if there's any change in the CLI.

vishalm0509: Ignore this comment for now - will get back on this.

server/internal/utils/telemetry/sync.go
```diff
@@ -13,18 +14,36 @@ import (
...
+	JobName            string
+	CreatedAt          time.Time
+	CreatedBy          string
+	SourceType         string
+	SourceName         string
+	SourceVersion      string
+	DestinationType    string
+	DestinationName    string
+	DestinationVersion string
+	Catalog            string
```
vishalm0509: Can we name this DestinationCatalog. As this term is also used for `streams` in the olake-cli.

server/internal/utils/telemetry/sync.go
```diff
@@ -35,6 +54,7 @@ func getJobDetails(jobID int) (*jobDetails, error) {
 	details := &jobDetails{
 		JobName:   job.Name,
 		CreatedAt: job.CreatedAt,
+		Catalog:   "unknown",
```
vishalm0509: ParseCatalogType sets it anyways right. We can remove it from here ?

ui/src/modules/ingestion/features/jobs/pages/Jobs.tsx
```diff
@@ -171,6 +177,10 @@ const Jobs: React.FC = () => {
 				A list of all your jobs stacked at one place for you to see
 			</p>
 
+			{jobs.some(job => job?.activate === true) && (
+				<CommunityHelpBanner source="jobs" />
+			)}
```
vishalm0509, line 182: Why are we showing this in presence of Active Jobs ?

server/internal/utils/telemetry/sync.go
```diff
@@ -46,59 +66,124 @@ func getJobDetails(jobID int) (*jobDetails, error) {
...
+
+		if details, err := getJobDetails(info.JobID); err == nil {
+			props["catalog"] = details.Catalog
 		}
+		return props, nil
 	}
 
-	if err := TrackEvent(ctx, eventType, properties); err != nil {
-		return err
+	// Old worker - upgraded workers always send Properties, so this path is dead post-upgrade.
```
vishalm0509: We can add a comment here that this is DEPRECATED

ui/src/modules/ingestion/features/jobs/pages/Jobs.tsx
deepanshupal09-datazip: earlier I added it only if recently a job failed... but then product asked to show it anyways so that more people can come to slack for help.

vishalm0509 (approved):
LGTM

PR #418, opened 2026-08-19
feat: s3-log-storage

saksham-datazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] normal sync and check commands working from ui
- [x] clear destination working from ui
- [x] Ran two jobs at once with heavy log load(10logs/second)
- [x] Backward compatibility for nfs via first running normal sync on staging for all three repos and than running them on my branch
- [x] Ran a sync and added commands such that worker oom happens in two times and their was no log loss
# Screenshots or Recordings

## Related PR's (If Any):
Olake helm=> https://github.com/datazip-inc/olake-helm/pull/189
Olake CLI => https://github.com/datazip-inc/olake/pull/1061


server/internal/storage/s3.go
```diff
@@ -0,0 +1,355 @@
...
+
+	if !customEndpoint {
+		_, err := client.HeadBucket(ctx, &s3.HeadBucketInput{Bucket: aws.String(bucket)})
+		if err != nil {
+			return fmt.Errorf("s3 bucket %q is not accessible: %s", bucket, err)
+		}
+		return nil
+	}
+
+	const maxAttempts = 60
```
vikaxsh, line 91: why 60

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,237 @@
...
+	return &dto.TaskLogsResponse{
+		Logs:         parseLines(lines),
+		OlderCursor:  currentFileNumber,
+		NewerCursor:  currentFileNumber + 1,
+		HasMoreOlder: currentFileNumber > firstFileNumber,
+		HasMoreNewer: currentFileNumber < lastFileNumber,
+	}, nil
+}
+
+func addS3FilesToArchive(ctx context.Context, workflowDir string, tarWriter *tar.Writer) error {
```
vikaxsh, line 183: we should keep the archive logs directory structure same as nfs, otherwise it would be difficult to read logs from chunked files
for exmaple 
logs/sync_*/connector-NNNNNN-*.log  -> logs/olake.log
logs/worker/worker-NNNNNN-*.log     ->  logs/worker.log

.gitignore
```diff
@@ -54,3 +54,8 @@ ui/test-results/
 ui/playwright-report/
 ui/blob-report/
 ui/tests/.auth/
+
+# ----------------------------
+# Local dev notes (not committed)
+# ----------------------------
+examples/
```
vikaxsh: why?

server/internal/utils/log_filesystem_reader.go
```diff
@@ -0,0 +1,387 @@
+package utils
```
vikaxsh, line 1: it would be better to use filesystem reader instead of nfs reader

.gitignore
saksham-datazip: Added newline and as for why I added the examples/ folder to .gitignore: I wanted to have a place to store commands or examples that are only meant for local development and shouldn’t be committed to the repository, similar to how we currently ignoring them in the CLI.

server/internal/storage/s3.go
saksham-datazip: It’s just a 5-minute maximum startup wait: 60 attempts × 5 seconds of sleep. If MinIO is already running, the first attempt succeeds and it returns immediately. The 60 attempts are only a worst-case cap, and we can use a different upper limit if needed.Also we already have the same function in worker as well

server/internal/utils/log_s3_reader.go
saksham-datazip: need to discuss

server/internal/utils/log_filesystem_reader.go
saksham-datazip: yeah changed the file name and some functions name related to it to log_filesystem_reader.go but didnt changed storagemode name.

server/internal/utils/log_s3_reader.go
saksham-datazip: Got it (nice idea) changed download logic such that only single file will be returned for each log instead of small files on download

server/internal/services/etl/destination.go
```diff
@@ -238,7 +238,8 @@ func (s Service) TestDestinationConnection(ctx context.Context, req *dto.Destina
 	homeDir := constants.DefaultConfigDir
 	mainLogDir := filepath.Join(homeDir, workflowID)
 	// Fetch the latest batch of logs by tailing from the end with default limit in the "older" direction.
-	logs, err := utils.ReadLogs(mainLogDir, -1, -1, "older")
+	// NOTE: the logs for test connection are fetched only for last present file in the logs directory.
```
vikaxsh, line 241: comment seems misleading

server/internal/utils/telemetry/telemetry.go
```diff
@@ -57,22 +60,35 @@ func InitTelemetry(db *database.Database) {
...
+			hash.Write([]byte(time.Now().String()))
+			newID := hex.EncodeToString(hash.Sum(nil))[:32]
 
-			if err != nil {
-				newID := func() string {
-					hash := sha256.New()
-					hash.Write([]byte(time.Now().String()))
-					return hex.EncodeToString(hash.Sum(nil))[:32]
-				}()
+			switch storagemode.Get() {
```
vikaxsh: Both switches could be grouped together

server/internal/appconfig/appconfig.go
```diff
@@ -47,6 +47,14 @@ type Config struct {
...
 	OptimizationPassword           string
 	OptimizationRequestTimeout     time.Duration
+	OlakeStorageMode               string
+	OlakeS3Bucket                  string
+	OlakeS3Region                  string
+	OlakeS3Prefix                  string
+	OlakeS3AccessKeyID             string
+	OlakeS3SecretAccessKey         string
+	OlakeS3SessionToken            string
+	OlakeS3Endpoint                string
```
vikaxsh, line 57: not used anywhere

vikaxsh: use appconfig for loading env where instead of directly calling viper.GetString in in s3.go and other files

server/internal/utils/log_s3_reader.go
vikaxsh: is it going to follow same folder structure as nfs?

server/internal/services/etl/destination.go
saksham-datazip: Changed

server/internal/appconfig/appconfig.go
saksham-datazip: done

saksham-datazip:
In olake-ui added versioning check such that if connector below v0.10.2 is selected than an error will pop up on the screen.

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	slices.SortFunc(chunks, func(a, b logChunkFile) int {
+		return a.index - b.index
+	})
+	return chunks, nil
+}
+
+// parseNumberedLogChunkName parses names like connector-000001-<timestamp>.log.
+func parseNumberedLogChunkName(name, prefix string) (int, bool) {
+	body := strings.TrimSuffix(name, ".log")
+	remainder := body[len(prefix)+1:]
```
mihir-datazip: `body[len(prefix)+1:]` has no length guard, and we never check the name actually starts with the prefix. `ConnectorLogPrefix` is "connector" so the index is always 10 — `olake.log` panics with `slice bounds out of range [10:5]`.

`ListObjectNames` hands us `path.Base` of every key under the sync dir, so a zero-byte directory marker (what the MinIO console or `aws s3 cp` on a dir leaves behind) gives `sync_2026-08-25_18-10-00`, which parses as chunk 8 and then gets fetched as a real chunk. `worker-000001-<ts>.log` also returns `(1, true)` and collides with the real connector chunk 1 under a non-stable `slices.SortFunc`.

gin.Recovery turns this into a 500 on every log poll for that job, permanently. Can we add a `HasPrefix` check and a length guard?

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func archiveNameUnderWorkflow(workflowDir, objectPath string) (string, bool) {
+	prefix := strings.TrimSuffix(workflowDir, "/") + "/"
+	name := strings.TrimPrefix(objectPath, prefix)
```
mihir-datazip: The NFS side anchors with `filepath.Join("logs", filepath.Base(filePath))`, but here we use the raw S3 key as the tar member name.

S3 keys are arbitrary strings and `keyToRelPath` doesn't Clean them, so a key containing `../` still passes the `HasPrefix(name, "logs/")` check, and Go's tar writer won't validate the name either. Any extractor that isn't hardened writes outside the target dir.

Can we keep the `filepath.Base` anchor here too?

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/datazip-inc/olake-ui/server/internal/utils/logger"
+)
+
+type logChunkFile struct {
+	index int
+	name  string
+}
+
+// readLogsFromS3 reads the logs from S3 for a workflow.
+func readLogsFromS3(ctx context.Context, workflowDir string, cursor int64, _ int, direction string) (*dto.TaskLogsResponse, error) {
```
mihir-datazip: `limit` is `_ int` here, so the API's limit param does nothing on the S3 path and we always return one whole chunk. The helm worker flushes at 1 MiB, so `?limit=10` comes back with a few thousand entries and ~1 MiB on the wire.

`TestSourceConnection` (`source.go:237`) and `TestDestinationConnection` (`destination.go:242`) both pass -1 relying on the NFS `limit <= 0 -> DefaultLogsLimit` clamp, which has no equivalent here.

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+		return err
+	}
+
+	objectPaths, err := storage.ListAllObjectRelPaths(ctx, path.Join(workflowDir, "logs"))
+	if err != nil {
+		return err
+	}
+
+	var (
+		connector, worker       bytes.Buffer
```
mihir-datazip: We `io.ReadAll` every chunk, convert to string, then copy into one `bytes.Buffer` that's only flushed after the loop — the NFS twin streams each file with `io.Copy`.

`bytes.Buffer` grows by doubling, so a job with 2 GB of chunks needs ~4 GB resident on a single request, and two concurrent downloads OOM the pod. `StreamLogArchive` describes itself as a streaming pipeline (`etl/job.go:725`), so I think we should stream here as well.

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+
+	for _, objectPath := range objectPaths {
+		archiveName, ok := archiveNameUnderWorkflow(workflowDir, objectPath)
+		if !ok {
+			continue
+		}
+
+		body, modTime, err := storage.ReadFileFromS3(ctx, "", objectPath, false)
+		if err != nil {
+			logger.Warnf("failed to add %s to archive: %s", objectPath, err)
```
mihir-datazip: A failed `GetObject` on one chunk only warns and continues. Since all connector chunks are concatenated into a single `logs/olake.log`, the gap lands in the middle of the log with no marker and no filename.

Headers and 200 are already committed by then, and the deferred closes still emit a structurally valid .tar.gz, so whoever debugs from that archive has no way to know anything was dropped. `addFilesystemFilesToArchive` aborts the whole archive on a per-file error — can we do the same here, or at least write a marker line?

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	// Read the lines from the chunk.
+	lines, err := readS3ChunkLines(ctx, syncLogDir, chunk)
+	if err != nil {
+		return nil, err
+	}
+
+	return &dto.TaskLogsResponse{
+		Logs:         parseLines(lines),
+		OlderCursor:  currentFileNumber,
+		NewerCursor:  currentFileNumber + 1,
```
mihir-datazip: `older_cursor`/`newer_cursor` are byte offsets in NFS mode and chunk indexes here, on the same untagged int64 field. NFS also clamps to file size (`log_filesystem_reader.go:283`) and this doesn't, and the handler does a bare `ParseInt` with no validation. The DTO example is `1704801600000`, so the Swagger contract is wrong for both.

Two concrete problems. `NewerCursor = index+1` means every line appended to the chunk we just served — up to 1 MiB of a running sync — is unreachable, where the NFS byte cursor re-reads exactly those bytes. And flipping the mode with a log tab open sends a byte offset as a chunk index: `newer` matches nothing and `taskStore.ts:202` writes back `has_more_newer=false` for the rest of that session.

server/internal/storage/s3.go
```diff
@@ -0,0 +1,358 @@
...
+
+// InitStorage initializes the shared S3 client when storage mode is S3. No-op for NFS.
+func InitStorage(ctx context.Context) error {
+	if storagemode.Get() != constants.StorageModeS3 {
+		return nil
+	}
+
+	cfg := appconfig.Load()
+	configOpts := []func(*config.LoadOptions) error{}
+	if cfg.OlakeS3Region != "" {
```
mihir-datazip: We hard-require the bucket but treat region as optional. aws-sdk-go-v2 still errors without a region even when `BaseEndpoint` is set — `endpoint rule error, A region must be set when sending requests to S3` (s3@v1.107.2/endpoints.go:3414).

So the natural MinIO setup (endpoint + bucket + keys, no region, since region is meaningless there) fails every call. And because `customEndpoint` is true, `ensureS3Bucket` then burns the full 60 x 5s loop before dying with a "failed to ensure s3 bucket" message that never names the real cause.

server/internal/storage/s3.go
```diff
@@ -0,0 +1,358 @@
...
+		var alreadyExists *s3types.BucketAlreadyExists
+		var alreadyOwned *s3types.BucketAlreadyOwnedByYou
+		if errors.As(err, &alreadyExists) || errors.As(err, &alreadyOwned) {
+			return nil
+		}
+
+		if attempt == maxAttempts {
+			return fmt.Errorf("failed to ensure s3 bucket %q after %d attempts: %s", bucket, maxAttempts, err)
+		}
+		time.Sleep(5 * time.Second)
```
mihir-datazip: This loop sleeps 5s x 60 with a bare `time.Sleep` and never checks `ctx.Done()`. `main.go:65` installs `signal.NotifyContext` (which disables Go's default SIGTERM disposition) and `main.go:68` calls `InitStorage` synchronously right after — so with a down endpoint we get ~295s where SIGTERM does nothing.

The compose healthcheck marks olake-ui unhealthy at ~105s, so temporal-worker and signup-init (both `service_healthy`) never start and `restart: unless-stopped` loops the stack. `utils.RetryWithBackoff` already exists (`utils/utils.go:117`) and is used at `temporal/client.go:37` — can we use it here?

server/internal/storage/s3.go
```diff
@@ -0,0 +1,358 @@
...
+	return ensureS3Bucket(ctx, s3Client, s3Bucket)
+}
+
+// ensureS3Bucket verifies the configured bucket exists. For S3-compatible endpoints
+// (MinIO), it creates the bucket when missing and retries while the server starts.
+func ensureS3Bucket(ctx context.Context, client *s3.Client, bucket string) error {
+	customEndpoint := appconfig.Load().OlakeS3Endpoint != ""
+
+	if !customEndpoint {
+		_, err := client.HeadBucket(ctx, &s3.HeadBucketInput{Bucket: aws.String(bucket)})
```
mihir-datazip: The retry policy looks inverted — dev/MinIO gets 5 minutes of retries and real AWS gets none. `if !customEndpoint` does a single `HeadBucket`, and any error goes to `logger.Fatalf` -> `os.Exit(1)`, so the `return` on `main.go:70` is dead and `defer stop()` never runs.

One transient 503 or an IMDS credential hiccup at pod start CrashLoopBackOffs the whole BFF, including every feature unrelated to S3. It also 403s under the natural least-privilege policy for a shared bucket: `s3:ListBucket` scoped with `Condition: StringLike s3:prefix` permits every ListObjectsV2 we issue but denies HeadBucket, which carries no prefix key.

server/internal/storage/s3.go
```diff
@@ -0,0 +1,358 @@
...
+	if s3Bucket == "" {
+		return fmt.Errorf("s3 bucket is required when storage mode is s3")
+	}
+	return ensureS3Bucket(ctx, s3Client, s3Bucket)
+}
+
+// ensureS3Bucket verifies the configured bucket exists. For S3-compatible endpoints
+// (MinIO), it creates the bucket when missing and retries while the server starts.
+func ensureS3Bucket(ctx context.Context, client *s3.Client, bucket string) error {
+	customEndpoint := appconfig.Load().OlakeS3Endpoint != ""
```
mihir-datazip: We're using "has a custom endpoint" as a proxy for "this is dev MinIO", but `OLAKE_S3_ENDPOINT` is set by every non-default-AWS deployment — R2, Ceph/RGW, on-prem MinIO in production, S3 VPC endpoints, LocalStack.

So a typo in `OLAKE_S3_BUCKET` silently creates a new empty bucket and writes all job configs, logs and state into it while ops watch the real one, with no error raised. `CreateBucketInput` also carries no `LocationConstraint`, so against a real-S3 endpoint outside us-east-1 it fails `InvalidLocationConstraint` on all 60 attempts.

We already provision buckets at the right layer — `server/tests/docker-compose.destination.yml` runs an `mc` sidecar doing exactly this. Can we drop the auto-create?

server/internal/storagemode/mode.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/datazip-inc/olake-ui/server/internal/constants"
+)
+
+// Get returns OLAKE_STORAGE_MODE from config, defaulting to nfs.
+func Get() string {
+	mode := appconfig.Load().OlakeStorageMode
+	if mode == "" {
+		return constants.StorageModeNFS
+	}
+	return mode
```
mihir-datazip: We return the env var verbatim with no case-normalization or validation, and all 11 dispatch points are `case StorageModeS3 / default`. So `OLAKE_STORAGE_MODE=S3` silently runs the entire server in NFS mode — `InitStorage` returns nil without creating a client, configs go to the pod's ephemeral `/tmp/olake-config`, `workerConfigPath` emits `/mnt/config/...` to workers with no shared mount, and log reads stat a directory that never exists.

The server starts clean with zero warnings and everything looks fine until the pod restarts — then all logs 404 and all sync state is gone. The bucket is already hard-required at `s3.go:72`, I think the mode deserves the same fail-fast.

server/internal/utils/telemetry/telemetry.go
```diff
@@ -57,22 +60,32 @@ func InitTelemetry(db *database.Database) {
...
+			case constants.StorageModeS3:
+				if idBytes, _, err := storage.ReadFileFromS3(ctx, "", relativePath, false); err == nil {
+					return string(idBytes)
+				}
+				id := newID()
+				_ = storage.WriteFilesToS3(ctx, constants.DefaultConfigDir, []storage.JobConfig{{RelativePath: relativePath, Data: id}})
+				return id
```
mihir-datazip: `err == nil` is the only success path, so any read failure — a 403 on GetObject with PutObject still allowed, a 503, a credential refresh between the two calls — falls through to `newID()` and then unconditionally PUTs over the canonical object, with the write error discarded too.

That permanently rotates the deployment's telemetry identity with nothing logged. The old path was `os.ReadFile` on a bind mount where ENOENT was the only realistic failure.

To tell the two apart we'd need `ReadFileFromS3` to wrap with `%w` first — all 12 `fmt.Errorf` in `storage/s3.go` use `%s`, so the SDK's typed NoSuchKey is already flattened to a string and `errors.As` can't work.

server/internal/utils/telemetry/sync.go
```diff
@@ -198,8 +200,7 @@ func enrichWithSyncStats(properties map[string]interface{}, workflowID string) e
 }
 
 func addStatsProperties(properties map[string]interface{}, mainSyncDir string) error {
-	statsPath := filepath.Join(mainSyncDir, "stats.json")
-	statsData, err := os.ReadFile(statsPath)
+	statsData, err := ReadSyncJobFile(context.Background(), mainSyncDir, "stats.json")
```
mihir-datazip: `ReadSyncJobFile` is a network call now, but we still pass `context.Background()` here and in `addStreamsProperties`. These run in fire-and-forget goroutines on every sync completion, so a hanging endpoint blocks each one for the retryer's 3 attempts plus TCP connect timeouts with no way to cancel, and the failures are only Debugf'd so it's invisible.

They also outlive main's signal context and hold up draining on SIGTERM. The request ctx is already threaded elsewhere in this PR — can we pass it down here?

server/internal/services/etl/metrics.go
```diff
@@ -272,11 +273,11 @@ func (m *metricsCollector) setStats(labels prometheus.Labels, stats *syncStats)
...
 
-// refreshStats re-reads stats.json for the last snapshot's jobs — filesystem only,
-// no DB or Temporal load. Caller holds m.mu.
-func (m *metricsCollector) refreshStats() {
+// refreshStats re-reads stats.json for the last snapshot's jobs (NFS or S3),
+// with no DB or Temporal load. Caller holds m.mu.
+func (m *metricsCollector) refreshStats(ctx context.Context) {
 	for _, target := range m.statsTargets {
-		stats, err := readSyncStats(m.statsBaseDir, target.workflowID)
+		stats, err := readSyncStats(ctx, m.statsBaseDir, target.workflowID)
```
mihir-datazip: `readSyncStats` is a blocking S3 GetObject now, and we do one per tracked job, sequentially, while holding `m.mu` (`metrics.go:130`), on the Prometheus scrape path with a 5s budget.

server/internal/utils/log_s3_reader.go
mihir-datazip: **F1 (high, correctness): unguarded slice — this panics, and it accepts objects that are not connector chunks.**

`body[len(prefix)+1:]` has no length guard, and the function never checks that the name actually starts with the prefix. `ConnectorLogPrefix` is `"connector"`, so the slice index is always 10. Running it verbatim:

- `"olake.log"` -> `slice bounds out of range [10:5]`
- `"a.log"` -> `[10:1]`
- `"logs"` -> `[10:4]`

**Failure scenario:** `ListObjectNames` returns `path.Base` of every key directly under `<hash>/logs/sync_*/`. That includes the zero-byte directory-marker object created by the MinIO/AWS console or by `aws s3 cp` on a directory, whose key equals the prefix — `path.Base` yields `sync_2026-08-25_18-10-00`, which parses as chunk index 8 and is then `GetObject`'d as a phantom chunk. `gin.Recovery` (`httpserver/server.go:35`) converts the panic into a 500 on **every** `GetTaskLogs` poll for that job, permanently.

Separately, `"worker-000001-<ts>.log"` returns `(1, true)` and collides with the real connector chunk 1 under a non-stable `slices.SortFunc`.

**Suggested fix:** require `strings.HasPrefix(body, prefix+"-")` and bounds-check before slicing; return `(0, false)` otherwise.

server/internal/utils/log_s3_reader.go
mihir-datazip, line 249: Got this review from AI, please verify once

**F2 (high, security): path traversal — this drops the `filepath.Base()` anchor its NFS twin keeps.**

`addFilesystemFilesToArchive` builds its member name as `archivePath := filepath.Join("logs", filepath.Base(filePath))` (`log_filesystem_reader.go:352`). This function uses the raw S3 key instead.

**Failure scenario:** S3 keys are arbitrary byte strings and `keyToRelPath` does not `Clean` them. An object at `<prefix>/<hash>/logs/../../../../etc/cron.d/x` — reachable in exactly the shared-bucket deployments `OLAKE_S3_PREFIX` exists for — survives the `strings.HasPrefix(name, "logs/")` check, and `addBytesToArchive` writes `tar.Header{Name: "logs/../../../../etc/cron.d/x"}`. Go's tar writer does not validate names. Any extractor of `job-<id>-logs-*.tar.gz` that is not itself hardened writes outside the target directory.

**Suggested fix:** mirror the NFS side — `filepath.Join("logs", filepath.Base(key))` — or reject any key whose cleaned relative path escapes the prefix.

server/internal/utils/log_s3_reader.go
mihir-datazip, line 26: Why is limit discarded here? If not needed we can remove from the function.

server/internal/utils/log_s3_reader.go
mihir-datazip, line 203: Each object goes `io.ReadAll` (`storage/s3.go:195`) -> `string` -> a single growing `bytes.Buffer`, which is only flushed after the loop completes (L234). The NFS twin streams each file with `io.Copy`.

Can you see if we can write each object's body straight into the tar writer as it is fetched, with the header size taken from the `ListObjects` metadata.

server/internal/utils/log_s3_reader.go
mihir-datazip, line 215: The NFS path aborts the whole archive on a per-file error: `return fmt.Errorf("failed to add files from logs directory %s: %s", logsDir, err)` (`log_filesystem_reader.go:352`) but here we warn and skip.

Is this intentional? It might silently drop chunks which is dangerous

server/internal/utils/log_s3_reader.go
mihir-datazip: **F12 (medium, API contract): the cursor means a byte offset in NFS mode and a chunk index here, on the same untagged `int64` field.**

NFS returns `linePos := readPos + int64(lastNL) + 1` and clamps it (`if cursor > fileSize { cursor = fileSize }`, `log_filesystem_reader.go:283`). This returns `int64(chunk.index)` with no clamp, and the handler does a bare `strconv.ParseInt` with no validation. The DTO example is `1704801600000`, a timestamp, so the Swagger contract is wrong for both backends.

**Failure scenario A — unreachable tail:** `NewerCursor = index+1` means every line appended to the chunk just served — up to 1 MiB of a running sync — cannot be addressed by any "newer" request. The NFS byte cursor re-reads exactly those bytes.

**Failure scenario B — mode flip:** an operator changes `OLAKE_STORAGE_MODE` with a log tab open, and the UI sends byte offset `152340` as a chunk index. `direction=older` silently serves the newest chunk from the wrong position; `direction=newer` matches nothing and returns the all-zero response, which `taskStore.ts:202-207` writes back as `has_more_newer=false` — so no newer fetch happens for the rest of that session.

**Suggested fix:** tag the cursor with its kind (or encode it as `"<kind>:<value>"`) and reject a cursor whose kind does not match the active mode, rather than misinterpreting it.

server/internal/storage/s3.go
mihir-datazip: **F8 (low): region is optional here but required by the SDK — helm covers this, nothing else does.**

The bucket is hard-required at L72; region is not. aws-sdk-go-v2's S3 endpoint ruleset errors without a region regardless of `BaseEndpoint` — verified in the pinned version at `aws-sdk-go-v2/service/s3@v1.107.2/endpoints.go:3414`.

olake-helm #189 makes this unreachable on the k8s path: `_helpers.tpl` `required`s both `s3LogFileStorage.bucket` and `.region`, and `olake.storageMode` fails the render without them. So this is only reachable from compose, a hand-rolled deployment, or anything consuming this package directly.

**Failure scenario (non-helm):** endpoint + bucket + static keys set, no `OLAKE_S3_REGION` / `AWS_REGION` — region being meaningless for MinIO. `config.LoadDefaultConfig` resolves an empty region and every request fails with `endpoint rule error, A region must be set when sending requests to S3.` Because `customEndpoint` is true, `ensureS3Bucket` then burns the full 60 x 5s loop, so the container hangs ~5 minutes and dies with a `failed to ensure s3 bucket` message that never names the real cause.

**Suggested fix:** cheap to close at this layer rather than relying on the chart — require region alongside bucket, or default to `us-east-1` when a custom endpoint is configured.

server/internal/storage/s3.go
mihir-datazip, line 111: Why the retry loop never checks `ctx.Done()?

I think we can also`utils.RetryWithBackoff`; it respects ctx.

server/internal/storage/s3.go
mihir-datazip: **F5 (medium, availability): the retry policy is inverted — real AWS gets zero retries, dev MinIO gets five minutes.**

`if !customEndpoint` performs a single `HeadBucket`. Any error returns, `main.go:69` calls `logger.Fatalf`, and `logger.go:99-102` is `logger.Fatal()...; os.Exit(1)` — so the `return` on `main.go:70` is dead and `defer stop()` never runs.

**Failure scenario:** one transient 503 `SlowDown`, or an STS/IMDS credential hiccup during pod start, `CrashLoopBackOff`s the entire BFF — taking down every feature, including everything unrelated to S3.

It also 403s under the natural least-privilege policy for a shared bucket: `s3:ListBucket` with `Condition: StringLike s3:prefix ["<OLAKE_S3_PREFIX>/*"]` permits every `ListObjectsV2` this code issues, but denies `HeadBucket`, which carries no `s3:prefix` key.

**Suggested fix:** retry both branches with the same bounded backoff, and prefer a prefix-scoped `ListObjectsV2` over `HeadBucket` as the reachability probe.

server/internal/storage/s3.go
mihir-datazip: **F7 (medium): `customEndpoint` is being used as a proxy for "this is dev MinIO", which it is not.**

`OLAKE_S3_ENDPOINT` is set by every non-default-AWS deployment: Cloudflare R2, Ceph/RGW, on-prem MinIO in production, an S3 VPC/gateway endpoint, LocalStack.

**Failure scenario:** a typo in `OLAKE_S3_BUCKET` silently creates a brand-new empty bucket and writes all job configs, logs and state into it, while ops watch the real one. No error is ever raised. Separately, `CreateBucketInput` carries no `CreateBucketConfiguration.LocationConstraint`, so against a real-S3 gateway endpoint in a non-`us-east-1` region it fails `InvalidLocationConstraint` on all 60 attempts.

**Do not simply delete the auto-create.** I checked olake-helm #189 — it ships the MinIO statefulset but never creates the bucket (no `defaultBuckets`, no `mc mb`, no create-bucket job), so this code is what makes the default `s3LogFileStorage.enabled: true` install work.

**Suggested fix:** gate creation on something that actually means "I own this bucket" — an explicit opt-in env var set by the chart only when it provisioned MinIO — and fail with a clear message otherwise. Alternatively create the bucket in the chart, the way `server/tests/docker-compose.destination.yml` already does with an `mc` sidecar (`if ! mc ls minio/warehouse; then mc mb minio/warehouse; fi`), and drop it here. Either way the two repos need to agree on who owns provisioning.

server/internal/storagemode/mode.go
mihir-datazip: **F3 (high, data loss): no case-normalization and no validation, so a typo silently runs the entire server in NFS mode.**

`Get()` returns `OLAKE_STORAGE_MODE` verbatim — `appconfig` only `TrimSpace`s it — and all 11 dispatch points are `case StorageModeS3 / default`.

**Failure scenario:** an operator sets `OLAKE_STORAGE_MODE=S3` (uppercase) with bucket and credentials. `"S3" != "s3"`, so: `InitStorage` returns nil at `s3.go:38` without creating a client or validating anything; `SetupConfigFiles` writes to the pod's ephemeral `/tmp/olake-config`; `workerConfigPath` emits `/mnt/config/...` to workers that have no shared mount; log reads `os.Stat` a directory that never exists. The server starts clean with **zero** warnings, and every job appears to work — until the pod restarts, at which point all logs 404 and all sync state is gone.

**The chart does not save you here.** olake-helm #189's `olake.storageMode` helper is `$mode := .Values.global.localStorageMode | default "nfs"` followed by `if eq $mode "s3"`, with no lowercasing. So `global.localStorageMode: S3` skips helm's own `required` checks on bucket and region *and* its Fusion guard, then ships `OLAKE_STORAGE_MODE: "S3"` straight into this function. A one-character values.yaml typo reaches the silent-fallback path through the supported install route.

**Suggested fix:** `strings.ToLower` the value and reject anything that is not a known mode. The bucket is already hard-required at `s3.go:72`; the mode deserves the same fail-fast. Worth normalizing in the chart helper too.

server/internal/appconfig/appconfig.go
```diff
@@ -121,5 +129,14 @@ func loadConfig() Config {
...
 		OptimizationPassword:       strings.TrimSpace(v.GetString("PASSWORD")),
 		OptimizationRequestTimeout: v.GetDuration("OPTIMIZATION_REQUEST_TIMEOUT"),
+
+		OlakeStorageMode:       strings.TrimSpace(v.GetString("OLAKE_STORAGE_MODE")),
+		OlakeS3Bucket:          strings.TrimSpace(v.GetString("OLAKE_S3_BUCKET")),
+		OlakeS3Region:          strings.TrimSpace(v.GetString("OLAKE_S3_REGION")),
+		OlakeS3Prefix:          strings.TrimSpace(v.GetString("OLAKE_S3_PREFIX")),
+		OlakeS3AccessKeyID:     strings.TrimSpace(v.GetString("OLAKE_S3_ACCESS_KEY_ID")),
+		OlakeS3SecretAccessKey: strings.TrimSpace(v.GetString("OLAKE_S3_SECRET_ACCESS_KEY")),
+		OlakeS3SessionToken:    strings.TrimSpace(v.GetString("OLAKE_S3_SESSION_TOKEN")),
```
mihir-datazip, line 139: Shouldn't we need these in `app.yaml`? If S3 mode is helm-only by design, say so in the README so the gap reads as intentional.

server/internal/utils/telemetry/telemetry.go
mihir-datazip: **F11 (medium): every read error is treated as "not found", then a fresh ID is PUT over the canonical object with the write error discarded.**

`if idBytes, _, err := storage.ReadFileFromS3(...); err == nil` is the only success path, so a 403 on `s3:GetObject` (with `PutObject` allowed), a 503 `SlowDown`, or a credential refresh between the two calls all fall through to `id := newID()` and `_ = storage.WriteFilesToS3(...)` — a plain `PutObject` with no `If-None-Match` and no existence re-check.

**Failure scenario:** the deployment's telemetry identity is permanently rotated, with nothing logged, on any transient read failure. The old path was `os.ReadFile` on a bind-mounted volume (`${PWD}/olake-data` -> `/tmp/olake-config`) whose only realistic failure was ENOENT, which is why "error means absent" held there and does not hold here.

**Suggested fix:** distinguish not-found from every other error before writing. That needs `ReadFileFromS3` to wrap with `%w` first — all 12 `fmt.Errorf` in `storage/s3.go` use `%s` and `grep -c '%w'` is 0, so the SDK's typed `NoSuchKey`/`NotFound` is already flattened to a string and `errors.As` is impossible.

server/internal/utils/telemetry/sync.go
mihir-datazip, line 203: Why context.Background() on network call?

server/internal/services/etl/metrics.go
mihir-datazip, line 280: `refresh()` holds `m.mu` for its whole duration (`metrics.go:130`) and re-runs `refreshStats` every `statsRefreshTTL = 2s`; `gather()` does the same per job in its loop. As `metricsRefreshTimeout` is 5s please verify that there are no scenarios where we reach timeout. In case this causes real issue, we can use an `errgroup` with `SetLimit`, and/or skip targets whose run already has a `CloseTime`

server/internal/utils/telemetry/sync.go
```diff
@@ -219,8 +220,7 @@ func addStatsProperties(properties map[string]interface{}, mainSyncDir string) e
 }
 
 func addStreamsProperties(properties map[string]interface{}, mainSyncDir string) error {
-	streamsPath := filepath.Join(mainSyncDir, "streams.json")
-	streamsData, err := os.ReadFile(streamsPath)
+	streamsData, err := ReadSyncJobFile(context.Background(), mainSyncDir, "streams.json")
```
mihir-datazip, line 223: Same here

server/internal/utils/telemetry/telemetry.go
```diff
@@ -57,22 +60,32 @@ func InitTelemetry(db *database.Database) {
...
-					hash := sha256.New()
-					hash.Write([]byte(time.Now().String()))
-					return hex.EncodeToString(hash.Sum(nil))[:32]
-				}()
+			switch storagemode.Get() {
+			case constants.StorageModeS3:
+				if idBytes, _, err := storage.ReadFileFromS3(ctx, "", relativePath, false); err == nil {
+					return string(idBytes)
+				}
+				id := newID()
```
mihir-datazip, line 75: Not all errors need to create new object, handle the 404 case and other errors gracefully

server/internal/utils/log_s3_reader.go
mihir-datazip: Each line is also JSON-unmarshalled twice — once in `isValidLogLine`, again in `parseLines`.

server/internal/utils/log_s3_reader.go
mihir-datazip, line 99: We should validate that name has prefix or at least enough length to do `body[len(prefix)+1:]` as it can cause panic

server/internal/utils/log_s3_reader.go
saksham-datazip: In our current codebase, this path is not possible because each workflow follows a fixed directory structure, and we do not add arbitrary files outside of that structure. Even if a pre-existing file with such a path were present in the bucket and ended up in the tar archive, it would simply be downloaded as part of the archive and would only help with debugging.

Additionally, the S3 bucket is dedicated to OLake, and the relevant workflow ID path is specific to that workflow.

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	index, err := strconv.Atoi(remainder[:dash])
+	if err != nil {
+		return 0, false
+	}
+
+	return index, true
+}
+
+// readS3ChunkLines reads the lines from a S3 log chunk.
+func readS3ChunkLines(ctx context.Context, syncLogDir string, chunk logChunkFile) ([]string, error) {
```
saksham-datazip:
```suggestion
// readS3ChunkLines reads and parses the valid, non-debug lines from a S3 log chunk.
func readS3ChunkLines(ctx context.Context, syncLogDir string, chunk logChunkFile) ([]LogEntry, error) {
```

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	chunkPath := path.Join(syncLogDir, chunk.name)
+	content, _, err := storage.ReadFileFromS3(ctx, "", chunkPath, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read log chunk %s: %s", chunkPath, err)
+	}
+
+	lines := make([]string, 0)
```
saksham-datazip:
```suggestion
	entries := make([]LogEntry, 0)
```

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+	for _, line := range strings.Split(content, "\n") {
+		line = strings.TrimSpace(line)
+		if isValidLogLine(line) {
+			lines = append(lines, line)
```
saksham-datazip:
```suggestion
		if entry, ok := parseValidLogLine(line); ok {
			entries = append(entries, entry)
```

server/internal/utils/log_s3_reader.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	return lines, nil
```
saksham-datazip:
```suggestion
	return entries, nil
```

server/internal/utils/log_utils.go
```diff
@@ -46,310 +90,62 @@ func isValidLogLine(line string) bool {
 	return true
```
saksham-datazip:
```suggestion
	return logEntry, true
```

server/internal/utils/log_utils.go
```diff
@@ -46,310 +90,62 @@ func isValidLogLine(line string) bool {
...
+	batch := make([]map[string]interface{}, 0, len(lines))
+	for _, line := range lines {
+		var logEntry LogEntry
 
-		// io.EOF is expected when reading near file boundaries
-		if rerr != nil && rerr != io.EOF {
-			return nil, 0, false, rerr
+		if err := json.Unmarshal([]byte(line), &logEntry); err != nil {
+			continue
 		}
```
saksham-datazip:
```suggestion
func parseLines(entries []LogEntry) []map[string]interface{} {
	batch := make([]map[string]interface{}, 0, len(entries))
	for _, logEntry := range entries {
```

server/internal/utils/log_s3_reader.go
saksham-datazip: Yes discarded limit from argument and we are doing it because for nfs it fetched by number of log lines for first fetch it will be 1000 lines and afterwards 500lines in each direction but in s3 we fetch by previous/next chunk so we dont need this parameter over here.
This was an existent behaviour on staging and this code previously existed in log_utils.go but now it is changed

server/internal/utils/log_s3_reader.go
saksham-datazip: Need to discuss

server/internal/appconfig/appconfig.go
saksham-datazip: s3 mode is for k8s mostly as docker does not need nfs specially for local but still because of storage and all user wants to configure s3 they will have to add it in docker-compose.yml by env variable we will further mention it in docs.

server/internal/utils/telemetry/sync.go
saksham-datazip: It was a leftover but fixed it now

server/internal/utils/log_s3_reader.go
saksham-datazip: i dont think ,this is possible as per our current flow but still added this check

server/internal/utils/telemetry/telemetry.go
saksham-datazip: It was their on staging as well even for hosted nfs so lets just keep it like that for now

server/internal/utils/telemetry/sync.go
saksham-datazip: Changed

server/internal/services/etl/metrics.go
saksham-datazip: Yes we can but this issue pre existed in staging so ig we can discuss it if we want to solve it now or add a todo?

server/internal/storage/s3.go
saksham-datazip: Added context select statement

server/internal/utils/log_s3_reader.go
saksham-datazip: In nfs mode we had very few files but here we might have files in gb meaning number of files can be in 1000s so i thought we can just skip while printing logs so as to get as much log as possible from user instead of failing the api and regarding the warn logs we have to ask from user saperately but we can `discuss` regarding this.

PR #419, opened 2026-08-21
feat: adding pos delete in streams

hash-data:
# Description
Added positional deletes in streams

image
image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


docker-compose.yml
```diff
@@ -104,6 +104,9 @@ services:
     environment:
       <<: *sharedEnvs
       OLAKE_CALLBACK_URL: "http://olake-ui:8000/internal/worker/callback"
+      OLAKE_INDEX_DB_DIR: /index-db
+      OLAKE_INDEX_DB_CACHE_SIZE: 512  # in mb
+      OLAKE_INDEX_DB_MAX_OPEN_FILES: 1000  # max open files
```
deepanshupal09-datazip, line 109: why are we adding index db to temporal worker?
is there any PR for adding index db to worker as well?

deepanshupal09-datazip: btw this file is deprecated now, we use `docker-compose-v1.yml` now.

ui/src/modules/ingestion/features/jobs/constants/streamConstants.ts
```diff
@@ -24,13 +26,33 @@ export const STREAM_DEFAULTS = {
...
+	{
+		label: "Positional",
+		value: UpsertType.POSITIONAL,
+		tooltip:
+			"Rows are deleted by their position in the existing data files before the new rows are written",
+	},
+]
+
+export const UPSERT_TYPE_TOOLTIP =
+	"Choose how existing rows are deleted when an upsert writes new versions of them"
```
deepanshupal09-datazip: please remove this... not used anywhere

ui/src/modules/ingestion/features/jobs/utils/streams.ts
```diff
@@ -75,9 +76,21 @@ export const getStreamsDataFromSourceStreamsResponse = (
...
+
 			mergedSelectedStreams[namespace].push({
-				...matchingSelectedStream,
+				...selectedStreamRest,
 				disabled: false,
+				// update_type only applies while the stream runs in upsert mode;
+				// older saved jobs carry no value, so fall back to the default.
+				...(upsertMode && {
+					update_type: savedUpdateType ?? DEFAULT_UPSERT_TYPE,
+				}),
```
deepanshupal09-datazip, line 113: since we are adding a new property in an existing selected stream, have we made sure that this will not trigger clear destination

ui/src/modules/ingestion/features/jobs/utils/streams.ts
```diff
@@ -94,6 +107,11 @@ export const getStreamsDataFromSourceStreamsResponse = (
 				stream_name: streamName,
 				disabled: true,
 				append_mode: !isDestUpsertModeSupported || !isSourceUpsertModeSupported, // Default to append if either source or destination does not support upsert
+				// update_type only applies while the stream runs in upsert mode.
+				...(isDestUpsertModeSupported &&
```
deepanshupal09-datazip, line 131: is there any min olake version required for this feature like column selection, filter_config etc. had?
right now we are adding this field unconditionally to even older jobs.

ui/src/modules/ingestion/features/jobs/constants/streamConstants.ts
```diff
@@ -24,13 +26,33 @@ export const STREAM_DEFAULTS = {
 	filter: "",
 } as const
 
+export const DEFAULT_UPSERT_TYPE = UpsertType.EQUALITY
```
deepanshupal09-datazip: we have `default_stream_properties` in `streams` for default values of normalization and append mode...
should this property's default be also derived from there instead of hardcoding in UI?

ui/src/modules/ingestion/features/jobs/utils/streams.ts
hash-data: yes

ui/src/modules/ingestion/features/jobs/utils/streams.ts
hash-data: yes there is a min version requirment but that will be written in docs, no failure will happen as this property will not be looked while unmarshalling the streams

ui/src/modules/ingestion/features/jobs/constants/streamConstants.ts
hash-data: we can do that let me check

docker-compose.yml
hash-data: On local testing i was using this one so added here, will add in v1.yaml. 
Also in worker this is being passed as env variable, worker directly share env variable with sync pods. It is required in sync pods. 


A Helm storage class pr will be separate

hash-data: One more thing I previously mentioned: no pr related to it will be on helm, but there is a pr as we need to pass the mounted folder to the olake CLI pod as well. Changes will be present in same pr of helm storage class

ui/src/modules/ingestion/common/types/streamTypes.ts
```diff
@@ -60,6 +65,9 @@ export type StreamData = {
 export interface DefaultStreamProperties {
 	normalization: boolean
 	append_mode: boolean
+	// Catalog-provided default upsert type. Stored on the selected stream as
+	// update_type. Absent on older olake versions.
+	update_mode?: UpsertType
```
deepanshupal09-datazip:
```suggestion
	update_type?: UpsertType
```
can we keep the naming consistent here?

deepanshupal09-datazip (approved):

PR #436, opened 2026-09-10
chore: upgrade Temporal from auto-setup 1.22.3 to server 1.32.0

vishalm0509:
# Description

- Replace deprecated `temporalio/auto-setup:1.22.3` with `temporalio/server:1.32.0`.
- Schema setup runs in `admin-tools` (`create` / `setup-schema` / `update-schema`) so existing Postgres volumes upgrade in place; namespace `default` is created if missing.
- UI and worker wait until that namespace exists. Install/upgrade docs download `scripts/` next to the compose file (bind-mounts; do not pipe compose from curl).

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Created multiple sources, dest, job, schedules in `auto-setup` version and checked if all persists after upgrade and data migration. 
- Checked all the Temporal based workflows in 

# Screenshots or Recordings

## Related PR's (If Any):

# Docs
- **Temporal Server**: https://datazip.atlassian.net/wiki/x/CQBWL
- **Temporal Admin Tools**: https://datazip.atlassian.net/wiki/x/AYBVL
- **Temporal UI**: https://datazip.atlassian.net/wiki/x/AgBVL


README.md
```diff
@@ -116,33 +125,40 @@ x-encryption:
 
 To update OLake UI to the latest version, use the command for your setup:
 
+```bash
```
vikaxsh: need to update in docs as well

docker-compose-v1.yml
```diff
@@ -151,30 +155,74 @@ services:
...
     environment:
       - DB=postgres12
       - DB_PORT=5432
       - POSTGRES_USER=temporal
       - POSTGRES_PWD=temporal
       - POSTGRES_SEEDS=postgresql
-      - TEMPORAL_ADDRESS=temporal:7233
-      - TEMPORAL_CLI_ADDRESS=temporal:7233
+      - SQL_PASSWORD=temporal
+    volumes:
```
vikaxsh: Use Compose top-level configs with inline content instead of scripts and keep every thing in one file only

docker-compose-v1.yml
```diff
@@ -151,30 +155,74 @@ services:
 
   temporal:
     container_name: temporal
-    image: ${CONTAINER_REGISTRY_BASE:-registry-1.docker.io}/temporalio/auto-setup:1.22.3
+    image: ${CONTAINER_REGISTRY_BASE:-registry-1.docker.io}/temporalio/server:1.32.0
```
vikaxsh, line 158: have we checked this?
https://docs.temporal.io/self-hosted-guide/upgrade-server#:~:text=Sequential%20Upgrades%3A%20Temporal%20Server%20should%20be%20upgraded%20sequentially%2C%20one%20minor%20version%20at%20a%20time.

scripts/temporal-create-namespace.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
vikaxsh: can we register default namespace from clients's code only instead of container and script?

vikaxsh:
Don't we need to upgrade for Legacy docker-compose.yml?

docker-compose-v1.yml
vishalm0509: I have checked this and have tested all the Temporal related processes that we have and it all works when I upgrade from old to new. 

- Schedule history exists
- Schedule runs
- Spec, Check, Frequency update, Cancel

All works fine. So for the features that we use of Temporal there are no breaking changes.

scripts/temporal-create-namespace.sh
vishalm0509: I have used the same way they have mentioned in the examples of temporal repo: https://github.com/temporalio/samples-server/blob/main/compose/docker-compose-postgres.yml

docker-compose-v1.yml
vishalm0509: I am moving and removing the dynamicconfig.

vishalm0509:
> Don't we need to upgrade for Legacy docker-compose.yml?

We are not maintaining this file anymore

docker-compose-v1.yml
```diff
@@ -151,30 +156,80 @@ services:
...
+    command: /scripts/create-namespace.sh
+    restart: on-failure:5
 
   temporal-ui:
     container_name: temporal-ui
-    image: ${CONTAINER_REGISTRY_BASE:-registry-1.docker.io}/temporalio/ui:2.16.2
-    profiles:
-      - debug
+    image: ${CONTAINER_REGISTRY_BASE:-registry-1.docker.io}/temporalio/ui:2.54.1
+    # profiles:
```
vikaxsh: why?

docker-compose-v1.yml
```diff
@@ -24,6 +24,7 @@ x-envs:
     OLAKE_SECRET_KEY: *encryptionKey
     PERSISTENT_DIR: *hostPersistencePath
     ENABLE_OPTIMIZATION: ${ENABLE_OPTIMIZATION:-false}
+    TEMPORAL_NAMESPACE: ${TEMPORAL_NAMESPACE:-default}
```
vikaxsh: if we are making this configurable, 
server/internal/services/etl/utils.go:252 still harcodes to default

docker-compose-v1.yml
vishalm0509: reverted. Commented out for testing purpose.

docker-compose-v1.yml
vishalm0509: Removed

vikaxsh (approved):

PR #439, opened 2026-09-11
feat: store target query engines and pass them to discover

vikaxsh:
# Description
Server support for the target query engines .

- `advanced_settings.target_query_engines` stores the engine selection on the job,
  the same path `max_discover_threads` takes. No migration — the column is jsonb.
- `StreamsRequest.target_query_engines` becomes `--target-query-engines` on discover,
  where OLake turns it into each stream's `available_update_types`.
- `available_query_engines` on the spec request returns the engine matrix the CLI
  ships, so the UI never carries its own copy.

Both flags are gated on `DefaultQueryEnginesVersion`; older images reject them as
unknown flags, so the spec call returns a null spec and the discover flag is omitted.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/server/docs/swagger.json
+++ b/server/docs/swagger.json
@@ -1877,7 +1877,7 @@
         },
         "/api/v1/project/{projectid}/sources/spec": {
             "post": {
-                "description": "Retrieve the UI spec for a specific source type/version.",
+                "description": "Retrieve the UI spec for a specific source type/version, or the supported query engines when available_query_engines is set.",
                 "tags": [
                     "Sources"
                 ],
@@ -2962,6 +2962,17 @@
                 "max_discover_threads": {
                     "type": "integer",
                     "example": 50
+                },
+                "target_query_engines": {
+                    "description": "TargetQueryEngines are the engines expected to read the destination tables. OLake\nkeeps them nowhere, so they are stored here and resent on every discover; changing\nthem requires regenerating the catalog to recompute available_update_types.",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "example": [
+                        "spark",
+                        "duckdb"
+                    ]
                 }
             }
         },
@@ -3655,6 +3666,11 @@
                 "version"
             ],
             "properties": {
+                "available_query_engines": {
+                    "description": "AvailableQueryEngines swaps the connector UI spec for the query engines the connector\nsupports and the delete formats each can read. The response carries a null spec on\nconnector versions that predate the feature.",
+                    "type": "boolean",
+                    "example": false
+                },
                 "type": {
                     "description": "enum: postgres,mongodb,mysql,mssql,db2,s3,kafka,iceberg",
                     "type": "string",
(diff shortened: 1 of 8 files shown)


server/docs/swagger.json
```diff
@@ -1877,7 +1877,7 @@
         },
         "/api/v1/project/{projectid}/sources/spec": {
             "post": {
-                "description": "Retrieve the UI spec for a specific source type/version.",
+                "description": "Retrieve the UI spec for a specific source type/version, or the supported query engines when available_query_engines is set.",
```
hash-data: it is generated via swagger command right?

server/internal/constants/constants.go
```diff
@@ -30,6 +30,8 @@ var (
 	DefaultSpecVersion               = "v0.2.0"
 	DefaultClearDestinationVersion   = "v0.3.0"
 	DefaultMaxDiscoverThreadsVersion = "v0.3.18"
+	// DefaultQueryEnginesVersion is the first release with the query engine flags; older images reject them.
+	DefaultQueryEnginesVersion = "v0.10.2"
```
hash-data:
```suggestion
	DefaultOlakeCLIQueryEnginesSupportVersion = "v0.10.2"
```

server/internal/services/etl/source.go
```diff
@@ -285,7 +286,7 @@ func (s Service) GetSourceVersions(ctx context.Context, sourceType string) (dto.
 
 // TODO: cache spec in db for each version
 func (s Service) GetSourceSpec(ctx context.Context, req *dto.SpecRequest) (dto.SpecResponse, error) {
-	specOut, err := s.temporal.GetDriverSpecs(ctx, "", req.Type, req.Version)
+	specOut, err := s.temporal.GetDriverSpecs(ctx, "", req.Type, req.Version, req.AvailableQueryEngines)
```
hash-data: why spec req will contain available query engines?

server/internal/services/temporal/execute.go
```diff
@@ -185,6 +199,11 @@ func (t *Temporal) GetDriverSpecs(ctx context.Context, destinationType, sourceTy
 	}, nil
 }
 
+// supportsQueryEngines reports whether the version takes the engine flags; custom builds are assumed current.
+func supportsQueryEngines(version string) bool {
```
hash-data: we might need to skip this check for dev mode

server/internal/constants/constants.go
vikaxsh: for other constants we are following this pattern only

server/internal/services/etl/source.go
vikaxsh: it is a boolean for passing flag in spec command

server/internal/services/temporal/execute.go
vikaxsh: with these envs in dev mode it should work
APP_ENV: development
 CUSTOM_DRIVER_VERSION: <version_tag>

server/docs/swagger.json
vikaxsh: yes

PR #441, opened 2026-09-15
fix: raise client timeouts for large stream catalogs

### LINKED ISSUE #434: Job creation/update fails with misleading timeout error on large sources
## Environment

- **Deploy method**: official `olake-helm` Helm chart (chart version `0.0.27`) on Kubernetes
- **Olake-UI version**: `v0.5.1`
- **OLake worker version**: `v0.3.10`
- **OLake connector version**: `v0.9.5`
- **OS**: Windows 11 (browser client); backend cluster nodes run Linux
- **Cloud Provider**: Self-hosted Kubernetes cluster, not a public cloud managed service
- **Docker Params**: N/A — deployed via Helm/Kubernetes rather than docker-compose

## Description

We had the problem that job creation and job updates in the OLake UI fail with a meaningless error message. When checking the browser console, we saw a timeout error from Axios, and the backend logs showed an `unexpected EOF` error. However, running the same requests directly via curl worked fine. The issue seems to be related to the timeouts in the UI, which are too short for certain operations like schema discovery. In our case, the source was a very large SQL Server database (Business Central) with many tables and columns.

## Steps to reproduce

1. Creating a new job failed. The UI showed a generic red error toast (something along the lines of "Something went wrong"), and the browser console showed:
   ```
   Error creating job: AxiosError: timeout of 10000ms exceeded
   ```
2. Running the same request directly via the terminal (curl) instead of through the browser worked fine. No timeout, request completed.
3. Afterward, editing that same job and saving an update to its stream config failed too, same kind of error, different threshold:
   ```
   Error updating job: AxiosError: timeout of 30000ms exceeded
   Error saving job: AxiosError: timeout of 30000ms exceeded
   ```
   Again worked via curl, no timeout, request completed.
4. We then checked the backend pod logs and saw a matching `unexpected EOF` around the same time as each failed browser request, e.g.:
   ```
   POST .../jobs/36/stream-difference     200   23.36s
   PUT  .../jobs/36                       400   33.16s   <- unexpected EOF
   ```
5. We tried again with a second job against the same source/destination some days later to reproduce the issue, and this time it worked without any timeout. Not sure why... Maybe we selected fewer streams or columns this time, but unsure if that would have made a difference.

**Browser Console Screenshot:**
Image

## Expected behavior

Job creation and updates should complete successfully, even if they take a bit longer for a large source like ours. Or at least fail with an error message that actually reflects what went wrong.

## Actual behavior

The UI shows a meaningless error toast and job is not created or updated, even though the same request completes fine via curl. The current 10s and 30s timeouts just seem too short for certain operations, especially with large sources.

krishanu-datazip:
# Description

Job creation and job updates fail in the browser with a meaningless error toast.
The console shows `AxiosError: timeout of 10000ms exceeded` (create) or
`30000ms exceeded` (update), while the identical request via `curl` succeeds.
The backend logs `unexpected EOF` and a `400` at the same moment.

**Cause:** an axios `timeout` covers connect + **request upload** + server
processing + response download, but these values were chosen to bound *server
processing* only. Four job endpoints carry the full stream catalog — every
discovered stream with its complete `type_schema` — which measures **9.08 MB**
for a source with 872 tables / 100,095 columns. On a slow uplink the upload
alone exhausts the budget, so axios aborts and the browser closes the
connection mid-body. `c.ShouldBindJSON` then hits end-of-stream, returns
`unexpected EOF`, and `StatusFromBindError` classifies it as a plain `400` —
written to a socket nobody is listening on. The user only ever sees the generic
toast.

**Change:** raise the budget on the four endpoints that move the catalog.

| Endpoint | Service method | Before | After |
|---|---|---|---|
| `GET /jobs/{id}` | `getJob` | 10s (inherited default) | 60s |
| `POST /jobs` | `createJob` | 10s (inherited default) | 60s |
| `PUT /jobs/{id}` | `updateJob` | 30s | 60s |
| `POST /jobs/{id}/stream-difference` | `getStreamDifference` | 30s | 0 |

`getStreamDifference` is left unbounded because, on top of the ~9 MB upload, its
handler runs a full discover workflow in a connector container server-side
(`discover --streams old --difference new`), so its duration is not bounded by
payload size alone.

`createJob` had no config object at all, so it silently inherited the 10s
default from `core/api/axios.ts` — it was the only job write without an explicit
timeout.

Also disables the **Create Job** button while the request is in flight and shows
an inline spinner, so a slow save cannot be submitted twice. Uses the `isPending`
flag already exposed by the existing `useCreateJob` TanStack mutation, and the
`SpinnerIcon` + `animate-spin` pattern already used in `tableColumns.tsx`.

> **This bounds the symptom, not the cause.** The payload still scales with
> source schema size, so a large enough catalog can still exceed 60s. The
> durable fix — making the payload proportional to what the user actually
> changed, rather than to how big the source is — is tracked separately.

Fixes #434

## Type of change

# How Has This Been Tested?

Reproduced end-to-end against a local stack pinned to the reported versions
(`olakego/ui:v0.5.1`, `olakego/ui-worker:v0.3.10`) with a seeded Postgres source
of **872 tables × 130 columns**, discovered through the real UI flow to produce a
genuine 9.08 MB catalog.

- [x] **Scenario A — the original failure reproduces.** With the upload
  rate-limited to 3G speeds, `POST /jobs` aborts at exactly 10.006s having
  uploaded only 983 KB of 11 MB; `stream-difference` and `PUT /jobs/{id}` abort
  at 30.005s. All three produce `unexpected EOF` + `400` server-side, matching
  the issue report verbatim:


ui/src/modules/ingestion/features/jobs/pages/JobCreation.tsx
```diff
@@ -413,11 +414,16 @@ const JobCreation: React.FC = () => {
...
 						onClick={handleNext}
+						disabled={isCreatingJob}
 					>
 						{currentStep === JOB_CREATION_STEPS.STREAMS ? "Create Job" : "Next"}
-						<ArrowRightIcon className="size-4 text-white" />
+						{isCreatingJob ? (
+							<SpinnerIcon className="size-4 animate-spin text-white" />
+						) : (
+							<ArrowRightIcon className="size-4 text-white" />
+						)}
```
deepanshupal09-datazip, line 429: here, shouldn't we also disable cancel, back and save job button?

deepanshupal09-datazip: also edit job has no loader, please add loader there as well.

ui/src/modules/ingestion/features/jobs/services/jobService.ts
```diff
@@ -39,6 +39,7 @@ export const jobService = {
 		try {
 			const response = await api.get<Job>(
 				`${API_CONFIG.ENDPOINTS.ETL.JOBS(API_CONFIG.PROJECT_ID)}/${id}`,
+				{ timeout: 60000 },
```
deepanshupal09-datazip: as discussed lets change timeout everywhere to 5 mins

ui/src/modules/ingestion/features/jobs/pages/JobCreation.tsx
deepanshupal09-datazip, line 306: this silently fails if streams.json is large and job is not saved.
do not need to fix it now, but we can create a github issue for this to pick up later.

deepanshupal09-datazip (dismissed):

ui/src/modules/ingestion/features/jobs/pages/JobCreation.tsx
krishanu-datazip: okay

ui/src/modules/ingestion/features/jobs/pages/JobEdit.tsx
```diff
@@ -66,6 +66,11 @@ const JobEdit: React.FC = () => {
 		JOB_CREATION_STEPS.STREAMS,
 	)
 	const [isSubmitting, setIsSubmitting] = useState(false)
+	// Tracks the stream-difference request, which runs before the update and can
+	// take a while on a large catalog. Kept separate from isSubmitting because
+	// handleJobSubmit runs nested inside handleStreamDifference on the no-diff
+	// path, and a shared flag would be cleared by the inner call.
+	const [isCheckingDifference, setIsCheckingDifference] = useState(false)
```
deepanshupal09-datazip: instead of creating and maintaining two separate local flags (isSumbitting, isCheckingDifference) we can use TanStack's `isPending` for both api calls, that makes it consistent with the rest of the code.

`updateJob` is already a tanstack mutation, so `isSubmitting` can be directly removed and use Tanstack's `isPending`
`getStreamDifference` - move this to a mutation inside `useJobMutations` and use it's `isPending`

remove both `isSubmitting` and `isCheckingDifference` state variables

deepanshupal09-datazip: Inside useJobMutations.ts, something like this:

```ts
export const useStreamDifference = () => {
	return useMutation({
		mutationFn: ({
			jobId,
			streamsConfig,
		}: {
			jobId: string
			streamsConfig: string
		}) => jobService.getStreamDifference(jobId, streamsConfig),
	})
}
```

and in JobEdit.tsx:

```ts
	const { mutateAsync: updateJob, isPending: isUpdating } = useUpdateJob()
	const { mutateAsync: getStreamDifference, isPending: isCheckingDifference } =
		useStreamDifference()
```

and now we directly use this mutation ln `handleStreamDifference` like we do in `handleJobSubmit`

ui/src/modules/ingestion/features/jobs/pages/JobCreation.tsx
deepanshupal09-datazip: please create a github issue for this.

deepanshupal09-datazip (approved):
LGTM

